PR #118, opened 2016-03-11
Removed "foreign key" property of "email" field in eventsignups

NotSpecial:
Not sure why it needed to be there anyways.

It seems to not break anything, but maybe we have some untested functionality here?

Hermann, you added this: Is this field necessary for something I cant see?

### DIFF
--- a/amivapi/models.py
+++ b/amivapi/models.py
@@ -438,7 +438,7 @@ class EventSignup(Base):
 
     event_id = Column(Integer, ForeignKey("events.id"), nullable=False)
     user_id = Column(Integer, ForeignKey("users.id"), nullable=False)
-    email = Column(CHAR(100), ForeignKey("users.email"))
+    email = Column(CHAR(100))
     additional_fields = Column(Text)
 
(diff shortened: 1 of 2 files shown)


hermannsblum:
According to the in-code documentation of the "email" field:

    'email': "For registered users, this is just a projection of your "
             "general email-address. External users need to provide their email"
             " here.",

the field should contain a projection of the user's email. If I remember correctly this is just for usability as you get with the corresponding rights and a GET to `/eventsignups` all the necessary contact details no matter if the users are external or internal.  

As far as I agree that the `ForeignKey` does not at all belong there we need an implementation of this projection or we change our policy.

NotSpecial:
We could solve this with a post get hook maybe, but I'm not sure if its necessary since you can just embed the user to get the mail address.

It would just make looping through the results easier since you always have to look at the same field.

But I was thinking about a different solution. What if we would provide a mailinglist for every event like we to for groups? Something like (eventid)@amiv.ethz.ch?

If you think that might be something I'll create an Issue^^

PR #121, opened 2016-03-31
Login with hook

### LINKED ISSUE #120: Changes to /sessions
There was some discussion between me an Fubu on the commit 1d3a88c0129e74e1808659a96ec4f9b71e89c355 wheter ETH-but-bot-member people should be able to log in.

I was against this and the commit was kept. Since then Conrad has raised a few points that made me reconsider this decision.

Some lectures are from ITET/MAVT, but are attended by students in MAT, INFK, etc.
But all documents for those lectures are in our database. Similar for the other way around: Lectures from e.g. INFK which are attended by our members.

For those reasons we have always shared those documents with all other students at ETH. So it probably should be possible for them to log in to access the documents. (Since we are sometimes not allowed to publish ETH documents to the public we need some authentication)

In combination with Fubus points (see commit discussion above) it seems necessary to allow non-members to log in again.

But on the same hand we have to reconsider our authentication levels to be aware of public/non-member/member differences, not only public/registered.

(But this should go into another issue)

On the other hand it would be nice to use eve methods for a post to /sessions instead of our own route. 

I think it should be possible to handle it similar to [this turoial](http://python-eve.org/tutorials/account_management.html#accounts-with-token-authentication)

Will try this later.

NotSpecial:
Closes #120 

All eth members can log in now.
Post to /session is now handled by Eve.

### DIFF
--- a/amivapi/authentication.py
+++ b/amivapi/authentication.py
@@ -4,8 +4,9 @@
 #          you to buy us beer if we meet and you like the software.
 
 """
-This file provides token based authentication (identification of users). A
-user can POST the /sessions resource to obtain a token.
+This file provides token based authentication (identification of users).
+
+A user can POST the /sessions resource to obtain a token.
 
 When a user sends his token with a request the g.logged_in_user global variable
 will be set.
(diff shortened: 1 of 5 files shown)


hermannsblum:
Do I understand correctly that the only thing this changes is the validation, which is now done by eve?

NotSpecial:
Not only, we do not have a custom route anymore.

Now the whole POST to /sessions is handled by eve like any other resource instead of our own route.

amivapi/authentication.py
```diff
@@ -91,7 +90,7 @@ def check_auth(self, token, allowed_roles, resource, method):
 authentication = Blueprint('authentication', __name__)
```
NotSpecial: This should probably be removed as well since its not needed anymore.

PR #123, opened 2016-04-14
Modularization

NotSpecial:
# Modularizing AMIVAPI

First of all I would recommend reading the docs for [flask extension development](http://flask.pocoo.org/docs/0.10/extensiondev/) and [application factories](http://flask.pocoo.org/docs/0.10/patterns/appfactories/). They are short and very informative, showing how extensions for flask can look and how they should be integrated in the app.

Short: Either have an extension class or module, in both ways there should be an `init_app` method and if you have a class you should avoid binding it to an app (E.g. `app.extension = Extension(app)` should not be done)

For this concept to work together with Eve we need to consider additional points: schema, validation, hooks. (Together with the flask blueprints)

## Schema

This is fortunately very easy since Eve provides a method `register_resource` which can be used just for this. We can just define the models and schema in the extensions and use this in `init_app`

The only difficulty are dependencies between sql models. The eve sqlalchemy `register_schema` relies on them, so all model classes need to be defined before `register_schema` is called. This can be handled easily by putting the schema creating in a method which is called in `init_app`. This way everything is imported before correctly.

(See _events.py_ and _groups.py_)

## Validation

Unlike for resources, there is no convenient method to extend the validator. On the other hand I didn't like having to manually combine it before app creation because I wanted to stick to the ``init_app`` concept.

Eve stores a validator **class** (not instance) in `app.validator`. I created a helper in _utils.py_ which creates a new, empty class with the extension class and the current validator class as parents. Using `type` this can be handled very nicely:

```python
app.validator = type("Adopted_%s" % validator_class.__name__,
                     (validator_class, app.validator),
                     {})
```

(More info in the [python docs](https://docs.python.org/2/library/functions.html#type))

## Hooks & Blueprints

No problem here, they can be added to the app instance in `init_app` comfortably.

# Status

Every resource (or related resources together) have now their own directory.
All resources are now modules.

_models.py_ and _schemas.py_ are gone


hermannsblum:
First of all: Nice work!  
I am convinced that this is really improving the chances that the api can be maintained easily and this project will last longer than out student IDs. 

Some comments:  

It is probably a question of personal style, but I don't like to put 1 module in 1 file. E.g. if you read through the events.py, it is quite long and not easy to understand how the data model, blueprints, hooks and validators work together and where you can find them. I would find it a more straightforward and ordered approach to have a stucture like than

    /events
        event_model.py
        event_validator.py
        event_hooks.py
        event_endpoints.py

However, I am not even sure if that is possible as I just read through the flask documentation and all these PR files. What do you think?

NotSpecial:
I see your point and I am not sure myself.

I'm afraid that more hierarchy like you propuse would actually make working with the code harder. At a first glance it might be more structured, but as soon as you are working with e.g. events you probably need to see all the code anyway.

So as soon as I want to work with events I now need to have four files open than just one.

But I really agree that its not looking super organized so far, but I like having all the code together.

Although I think it might get better after swichting to mongo. Now the resource module looks like this:
```
several classes for db (okay)
function to create domain from db (quite ugly)
class for validator (okay)
hook function (sometimes many of them, but not that bad looking)
init function
```
Using vanilla Eve (with mongo) this would be just
```
eve schema, one big dict (pretty)
validator class (okay)
hooks (okay)
init function
```

I think this would be organized enough.
I'm afraid that by splitting up the modules we will need lots of imports everywhere which will be confusing to get an overview and make it harder to change something.

Maybe @fubu could add something here as a more experienced python programmer.

NotSpecial:
its done, all resources have been put into modules. Updating status in first comment.

hermannsblum:
Looks good to me

PR #133, opened 2016-11-12
Switch to Mongo complete. Merge back into Master

NotSpecial:
It is time, we are finally done switching to Mongo and have basically rewritten most of the API in the progress.

Now all functions are included again so we can finally switch back to master.


NotSpecial:
Doesn anybody have an idea why travis fails?

All tests get a mongo error: `OperationFailure: auth fails`

Things I have ruled out so far:
- config file can't be found: Not the case, tried this locally and that produces a different error since config loading happens before first mongo request.
- mongo refusing connection: both localhost and 127.0.0.1 explicitly stated fail

marcoep:
Also fix issue  #134 please before merging.

NotSpecial:
I am wonderinf if we should exclude the ldap files from the coverage.
(Would be easily achievable with a filter for coveralls)

Pro:
- We can't test ldap on travis because we can't publish the credentials (so its always nearly no coverage)

Contra:
- It might get forgotten that you have to take care of ldap tests too.

Or, in other words: Does anybody have an idea if we can test ldap on travis somehow?

Tar-Tarus:
I think one essential element for __studydocs__ is missing: **the semester in which the document has been composed** (aka __"Jahr"__ as it currently appears on the website). As a first-year student relying heavily on AMIV's studydocs repository, it was on of the main search criteria I used because it is simply related to the proximity of the actual content of the course I had.
Can anyone provide me with a valid reason why this has been omited? Otherwise, I would quickly add it as a required field posting a file.

amivapi/auth/link_methods.py
```diff
@@ -0,0 +1,185 @@
...
+
+from flask import current_app, g
+from eve.auth import resource_auth
+
+from .auth import AmivTokenAuth, authenticate, check_if_admin
+
+
+def _get_item_methods(resource, item):
+    res = current_app.config['DOMAIN'][resource]
+    user = g.get('current_user')  # TODO: post_internal_problem
```
Tar-Tarus: What is the issue concretely?

amivapi/studydocs/endpoints.py
```diff
@@ -80,6 +81,11 @@
                 'maxlength': 30,
                 'nullable': True,
                 'type': 'string'
+            },
+            'coursesemester': {
+                    'maxlength': 5,
+                    'nullable': True,
+                    'type': 'string'
```
NotSpecial: I am not sure if we gain anything by this. All items already have a creation date - so for proximity you could just sort by that.

Am I missing something?

amivapi/tests/ldap_integration.py
```diff
@@ -6,194 +6,258 @@
...
-There set them first:
-
+Since we don't have access to test accounts for LDAP you need to provide
+valid ETH data for the tests to work.
+
+This means we need lots of environment variables:
+- ldap_nethz
+- ldap_password
+- ldap_firstname
+- ldap_firstname
```
Tar-Tarus: Why is __ldap_firstname__ appearing twice?

NotSpecial: Seems like a typo, can be removed

amivapi/auth/link_methods.py
NotSpecial: Good catch! This todo can go away, since its ok like this. At the time I wasn't sure, but there is not always a user so using a `.get()` here is justified. (e.g. using an apikey)

But I am a little unhappy with how the link methods look. I have an idea how to code it nicer and have created an issue for this. But its not changing the result so i think its not critical.

amivapi/studydocs/endpoints.py
Tar-Tarus: Well the idea is that students can upload their documents at any time and link it to the semester where they had taken the course and had made this document. This can happen during the semester as much as it can happen when a Master student digging through his old documents finds a document which might be helpful and decides to upload it.

NotSpecial: Ok, I see your point. But then we maybe should use 'courseyear' or something and validate its a year (e.g. with regex) - I think this is important to ensure that this field is comparable accross several items.

E.g. the first person could write: FS12, the next person 2014, etc - this makes it very complicated for a tool (e.g. the website) to parse it. 

I have created an issue for this: #136

NotSpecial: Ah, i can see your point, but then we need to add more validation, see my other comment

Tar-Tarus: Yes that was definitely my intent behind it. Was a bit unclear in my head as to how I would do it but Regex validation seems right.

amivapi/auth/link_methods.py
```diff
@@ -0,0 +1,185 @@
...
+    if 'parent' in links:
+        links['parent']['methods'] = _home_link_methods()
+
+    # collection -> resource (optional)
+    if 'collection' in links:
+        links['collection']['methods'] = _get_resource_methods(resource)
+
+
+def add_methods_to_resource_links(resource, response):
+    """Add methods to all links of the item.
```
hermannsblum: This looks like a copy-paste ;)
It is unclear to me where the difference to the above method lays. 
If I understand corrrectly, this here only adds links to a response?

amivapi/auth/link_methods.py
```diff
@@ -0,0 +1,185 @@
+# -*- coding: utf-8 -*-
+#
+# license: AGPLv3, see LICENSE for details. In addition we strongly encourage
+#          you to buy us beer if we meet and you like the software.
+
+"""Functions adding permitted methods to '_links' part of response.
```
hermannsblum: the filename is quite generic, it took me some time to figure out what you are dooing here.
Something like authorized_link_methods would probably be more specific.

amivapi/auth/link_methods.py
```diff
@@ -0,0 +1,185 @@
...
+
+def _get_data(response):
+    """Get response data as dict.
+
+    Encoding/Decoding necessary for compatibility with both python 2 and 3.
+    """
+    return json.loads(response.get_data().decode('utf-8'))
+
+
+def _set_data(response, data):
```
hermannsblum: these two methods seem oddly displaced just in between all the other code

amivapi/auth/link_methods.py
```diff
@@ -0,0 +1,185 @@
...
+            pass
+        else:
+            # Add links for home
+            for res_link in links:
+                res_name = res_link['title']  # title equals resource
+                if isinstance(resource_auth(res_name), AmivTokenAuth):
+                    check_if_admin(res_name)
+                    res_link['methods'] = _get_resource_methods(res_name)
+
+            _set_data(response, data)
```
hermannsblum: After going through this file, I wonder if there is any possibility to write this more clearly. All the different methods are really confusing and it is hard to get an overview if at some point somebody wants to change comething. But as long as it works I would simply add this as issue 😉

amivapi/auth/sessions.py
```diff
@@ -0,0 +1,204 @@
...
+                'POST': "Login and aquire a login token. Post the fields "
+                "'username' and 'password', the response will contain the "
+                "token. username can be either nethz, mail, or user_id",
+                'GET': "Check token(s)."}
+        },
+
+        'authentication': SessionAuth,
+        'public_methods': ['POST'],
+        'public_item_methods': [],
+        'resource_methods': ['GET', 'POST', 'DELETE'],
```
hermannsblum: Why do you allow DELETE on the resource?

amivapi/auth/sessions.py
```diff
@@ -0,0 +1,204 @@
...
+        # update password - hook will handle hashing
+        update = {'password': plaintext}
+        with admin_permissions():
+            patch_internal("users", payload=update, _id=user['_id'])
+    return is_valid
+
+
+# Regular task to clean up expired sessions
+@periodic(datetime.timedelta(days=1))
+def delete_expired_sessions():
```
hermannsblum: For clarification: as the periodicity is set to 1 per day, if somebody sets a session timeout of 15 mins, it may be that the session only times out after 1 day and 15 mins.
Is this intended?

amivapi/auth/sessions.py
```diff
@@ -0,0 +1,204 @@
...
+    for item in items:  # TODO (Alex): Batch POST doesnt really make sense
+        username = item['username']
+        password = item['password']
+
+        # LDAP
+        if (config.ENABLE_LDAP and
+                ldap_connector.authenticate_user(username, password)):
+            # Success, sync user and get token
+            updated = ldap_connector.sync_one(username)
+            _prepare_token(item, updated['_id'])
```
NotSpecial: _id is how mongo and eve use the field. Eve sqlalchemy changed this to id. Now it's_id again because eve can handle this without modifications.

amivapi/auth/auth.py
```diff
@@ -0,0 +1,321 @@
...
+4. Tip: You can still use all Eve settings like 'public_methods' and
+   'public_item_methods'!
+
+Take a look at `users.security` for an example.
+
+
+**How it works**
+
+We are using Eve's hooks to add authentiation.
+All methods go through the following methods:
```
hermannsblum: All requests?

amivapi/auth/sessions.py
NotSpecial: Very good question, this should not be allowed

amivapi/auth/sessions.py
```diff
@@ -0,0 +1,204 @@
+# -*- coding: utf-8 -*-
+#
+# license: AGPLv3, see LICENSE for details. In addition we strongly encourage
+#          you to buy us beer if we meet and you like the software.
+"""Sessions endpoint."""
```
hermannsblum: Very well documented, can read it fluently 👍 😊

amivapi/auth/auth.py
```diff
@@ -0,0 +1,321 @@
+# -*- coding: utf-8 -*-
+#
+# license: AGPLv3, see LICENSE for details. In addition we strongly encourage
+#          you to buy us beer if we meet and you like the software.
+
+"""Custom AMIV Token auth class.
+
+**FAQ**
```
hermannsblum: ❤️

amivapi/auth/auth.py
```diff
@@ -0,0 +1,321 @@
...
+    """Call function only if `g.get('auth_required')`."""
+    @wraps(func)
+    def wrapped(*args):
+        if g.get('auth_required'):
+            func(*args)
+    return wrapped
+
+
+def not_if_admin(func):
+    """Call function only if `g.resource_admin`."""
```
hermannsblum: I guess a NOT is missing here

amivapi/auth/auth.py
NotSpecial: Indeed. (Except GET on home,  which is handled a bit special by eve,  but all requests to api resources and items.)

amivapi/events/endpoints.py
```diff
@@ -8,184 +8,253 @@
...
+                'than their email-address',
+                'allow_email_signup': 'If False, only AMIV-Members can sign '
+                'up for this event',
+                'price': 'Price of the event as Integer in Rappen.',
+                'spots': "For no limit, set to '0'. If no signup required, "
+                "set to '-1'. Otherwise just provide an integer."
+            },
+            'general': 'An Event is basically everything happening in the '
+            'AMIV. All time fields have the format YYYY-MM-DDThh:mmZ, e.g. '
+            '2014-12-20T11:50:06Z',
```
hermannsblum: didn't this change for JS reasons? Anyway, the example and the description are not consistent

Tar-Tarus:
Checkout #139

amivapi/events/endpoints.py
NotSpecial: I created a new issue for this: #140

NotSpecial:
This is all complete. But we cannot rebase this branch to master without problems. I have created a new branch that is rebased to master (after some struggles). See #142

PR #145, opened 2016-12-10
Swagger rebased updated

NotSpecial:
Fancy new swagger interface for /docs added by hermann


hermannsblum:
Seems like there are still some rebase problems, travis fails...

amivapi/auth/sessions.py
```diff
@@ -38,16 +38,14 @@ def create_user_lookup_filter(self, user_id):
...
-                'GET': "Check token(s)."}
-        },
+        'description': "A session is used to authenticate a user after he "
+        "provided login data. A POST to /session will return a token you can "
+        "use in an Authorization header: token <yourtoken>",
+        # 'methods': {
+        #    'POST': "Login and aquire a login token. Post the fields "
+        #    "'username' and 'password', the response will contain the "
+        #    "token. 'username' can be either nethz, mail, or user_id",
+        #    'GET': "Check token(s)."}
```
Leonidaz0r: Is this documentation lost now? If yes, can we add it somewhere else?
If no, can we remove this comment?

amivapi/docs.py
```diff
@@ -0,0 +1,62 @@
...
+    app.register_blueprint(swagger_ui)
+
+    register_validator(app, DocValidator)
+
+    # replace null-type fields with ''
+    # TODO: make this dynamic
+    add_documentation({'definitions': {'User': {'properties': {
+        'password': {'default': ''},
+        'nethz': {'default': ''}
+    }}}})
```
Leonidaz0r: What is the purpose of this? We generate empty documentation fields?

amivapi/docs.py
```diff
@@ -0,0 +1,62 @@
...
+
+    add_documentation({'securityDefinitions': {
+        'AMIVauth': {
+            'type': 'apiKey',
+            'name': 'Authorization',
+            'in': 'header',
+            'description': 'enter a token you got with POST to /sessions'
+        }
```
Leonidaz0r: This is badly worded, as it suggests API keys and tokens are the same. Looking at the created website, I propose:

name: Authorization header
remove in
description: 'enter either a session token, you retrived using POST to /sessions, or an API key defined in the server config.'

amivapi/events/model.py
```diff
@@ -14,27 +14,12 @@
...
-                'to view AMIV-Events'
-            }
-        },
-
+        'description': 'An Event is basically everything happening in the '
+        'AMIV. All time fields have the format YYYY-MM-DDThh:mmZ, e.g. '
+        '2014-12-20T11:50:06Z',
+        # 'methods': {
+        #     'GET': 'You are always allowed, even without session, '
+        #     'to view AMIV-Events'
```
Leonidaz0r: Again: do we loose this? remove comment?

amivapi/events/model.py
```diff
@@ -173,24 +166,12 @@
...
-                'PATCH': 'Only additional_fields can be changed'
-            }
-        },
-
+        'description': 'You can signup here for an existing event inside of '
+        'the registration-window. External Users can only sign up to public '
+        'events.',
+        # 'methods': {
+        #     'PATCH': 'Only additional_fields can be changed'
+        # }
```
Leonidaz0r: Same thing again

amivapi/groups/model.py
```diff
@@ -47,15 +47,15 @@ class GroupMembershipAuth(AmivTokenAuth):
     """Auth for group memberships."""
 
     def has_resource_write_permission(self, user_id):
-        """All user can signup for groups.
+        """All user can enroll into groups.
```
Leonidaz0r: enroll stands with for, in or on, not with into
https://dictionary.cambridge.org/dictionary/english/enrol?a=british

amivapi/settings.py
```diff
@@ -116,3 +116,37 @@
     'mtec',
     'gess'
 ]
+
+# Config for swagger (/docs)
+# TODO: should be linked to the setup.py information in the package
+SWAGGER_INFO = {
+    'title': 'AMIVAPI',
```
Leonidaz0r: I would write "AMIVAPI Documentation", as that word is missing on the page so far. However this is just an opinion.

amivapi/settings.py
```diff
@@ -116,3 +116,37 @@
...
+    'version': '0.1-dev',
```
Leonidaz0r: We should get this from a constant somewhere to make it automatically update over releases.

amivapi/settings.py
```diff
@@ -116,3 +116,37 @@
...
+    'description': "The REST API behind most of AMIV's web services.",
+    'termsOfService': 'todo',
+    'contact': {
+        'name': 'AMIV an der ETH',
+        'url': 'http://amiv.ch/'
```
Leonidaz0r: https please

amivapi/settings.py
```diff
@@ -116,3 +116,37 @@
...
     'gess'
 ]
+
+# Config for swagger (/docs)
+# TODO: should be linked to the setup.py information in the package
+SWAGGER_INFO = {
+    'title': 'AMIVAPI',
+    'version': '0.1-dev',
+    'description': "The REST API behind most of AMIV's web services.",
+    'termsOfService': 'todo',
```
Leonidaz0r: Add a # TODO comment, so this shows up in grep

amivapi/settings.py
```diff
@@ -116,3 +116,37 @@
...
+    'contact': {
+        'name': 'AMIV an der ETH',
+        'url': 'http://amiv.ch/'
+    },
+    'license': {
+        'name': 'GNU',
```
Leonidaz0r: This just puts "GNU" without context to the website. We need to write what it is.
Change to "GNU Affero General Public License"?

amivapi/settings.py
```diff
@@ -116,3 +116,37 @@
...
+                       'abort_if_not_public',
+                       'add_lookup_filter',
+                       'check_resource_write_permission',
+                       'check_item_write_permission',
+                       'add_permitted_methods_after_insert'
+                       'add_permitted_methods_after_fetch_resource',
+                       'add_permitted_methods_after_fetch_item',
+                       'add_permitted_methods_after_fetch_resource',
+                       'add_permitted_methods_for_home',
+                       'add_permitted_methods_after_update']
```
Leonidaz0r: This does not seem to work. add_permitted_methods_after_insert is shown under POST to joboffers.
Any ideas?

amivapi/settings.py
```diff
@@ -116,3 +116,37 @@
...
+X_DOMAINS = ['http://127.0.0.1:5000',  # The domain where Swagger UI is running
```
Leonidaz0r: This IP should not be hardcoded, this breaks the documentation in many setups. The address should be available via flask. (maybe SERVER_NAME config value?)

Leonidaz0r:
I fixed all the comments I made apart from the one about the code in docs.py line 50. It is not clear to me, what that code is for.

Still design does not look good and we are missing the validator information.

Leonidaz0r:
Looking at other swagger based documentation, I think we should move the description of resources to implementation_notes and just have a short string for the description.

amivapi/docs.py
hermannsblum: Swagger in the currrent version does not support

    Default: null

And crashes. They decided to fix this only in later versions, so for now we have to replace all null fields with something else

amivapi/docs.py
hermannsblum: This is swagger code. Swagger supports 3 types of authorization schemes, and they call token-based authorization 'apiKey'

'name' is the key of the header field, this cant be changes therefore.

But if course we can adapt the description field

amivapi/settings.py
hermannsblum: I discussed this with @NotSpecial

It basically is the same information as in the package-info. However, we did find bith directions of import weird, so we left it like this for the moment.

Do you have a suggestion how to do it?

hermannsblum:
I fixed the authorization again following my comments above

hermannsblum:
I think the documentation with realtion to special HTTP methods is OK now, even though it should in the end only be displayed for the actual method you look at right now.

Also related to the more detailed description and "summary": How do you suggest to structure the json? We could think of something like this:

        'events': {
            'summary': 'Anything happening in the AMIV'.
            'description': 'An Event is basically everything happening in the '
            'AMIV. All time fields have the format YYYY-MM-DDThh:mmZ, e.g. '
            '2014-12-20T11:50:06Z\n\n'
            'method_documentation' {
                'GET': 'This is public even without a session'
            }

We would then adapt the mapping into the swagger fields, this is done by eve-swagger and therefore easily adaptable as we already use our own fork

amivapi/settings.py
fubu: The way many popular packages do it is to include a `__version__ = '0.1-dev'` in `amivapi/__init__.py` and read that from `setup.py` via a regular expression (like `r'__version__\s+=\s+(.*)'`).

Examples:
* [flask](https://github.com/pallets/flask/blob/master/setup.py#L52-L56)
* [requests](https://github.com/kennethreitz/requests/blob/master/setup.py#L52-L54)
* [SQLAlchemy](https://github.com/zzzeek/sqlalchemy/blob/master/setup.py#L112-L118)

There are also [other possibilities](https://packaging.python.org/single_source_version/), but IMHO the above approach works best. (See the bullet points in [this StackOverflow answer](https://stackoverflow.com/a/7071358) for some reasons why to use regular expressions for this task.)

Leonidaz0r:
I like that approach.

hermannsblum:
For the documentation of validation rules: This is a [known swagger issue](https://github.com/swagger-api/swagger-ui/issues/1217) right now.

It looks like they lost this with the implementation of the example objects as they had it before. Apparently you can see some stuff by hovering, but I do not consider this a solution.

As we can expect this issue to be worked on, I suggest an easy solution for the moment printing out the schema into the implementation notes via our version of eve-swagger.

hermannsblum:
Doing some more research 😃

I think I actually prefer this much more over swagger: [http://rebilly.github.io/ReDoc/](http://rebilly.github.io/ReDoc/)  
It uses the same documentation resource (swagger-like JSON description is basis of OpenAPI), but describes actually all the stuff we want.
However, it does not yet have an interactive console to try out requests like swagger does...

What are your opinions?

Leonidaz0r:
I have at least two issues:
1. import error: bootstrap.py imports docs, but the file is called documentation.py
2. I don't like the structure. swagger_ui as a folder looks like it is part of the API, while it is not. Actually there are three things in the amivapi package, which all belong together (swagger_ui folder, documentation.py and ReDoc folder). The swagger_ui folder looks like it is API code, while actually it is a library. I would recommend we either have a folder "lib" or "external" or whatever and put it in there, or we move it outside of amivapi altogether and somehow reference it.
An easy solution to 2 might be to have:

amivapi/documentation/lib/swagger_ui/
amivapi/documentation/ReDoc/
amivapi/documentation/documentation.py

hermannsblum:
You fixed the import error by yourself, thanks and sorry for that 🤙

Concerning the structure: This was actually only there as I never really got a feedback whether we should use ReDoc or SwaggerUI. As nobody seems to care too much, I simply decided for ReDoc, which we can include as external JS and therefore we only need the file `amivapi/documentation.py` which now inlcudes the 14 lines HTML template.

PR #151, opened 2016-12-11
Added show_website flag to joboffers and reworked other fields

Leonidaz0r:
We discussed the requirements of job offers.
It was decided, that any joboffer will always have a pdf and a company logo, therefore we can make them required. This makes integration with CRM and displaying more reliable as well.

In addition we want a flag, that controls whether the job offer is shown on the website. This is important, as some companies want to take a job offer offline, when the job is given to someone. In addition we would like a possibility in the future, where companies can create joboffers themselves on the kontakt page. That means we want to review them before publishing, which we can, if the flag can prevent it from being shown.

### DIFF
--- a/amivapi/tests/fixtures.py
+++ b/amivapi/tests/fixtures.py
@@ -273,6 +273,9 @@ class FixtureMixin(object):
         obj.setdefault(
             'title_de',
             self.create_random_value(schema['title_de']))
+        obj.setdefault(
+            'description_de',
+            self.create_random_value(schema['description_de']))
 
     def create_random_value(self, definition):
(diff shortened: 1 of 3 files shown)


amivapi/tests/fixtures.py
```diff
@@ -329,7 +332,7 @@ def create_random_value(self, definition):
 
         elif t == 'media':
             ftype = random.choice(definition.get('filetype', ['zip']))
-            if ftype == 'jpg':
+            if ftype == 'jpg' or ftype == 'jpeg':
```
fubu: better: `if ftype in ('jpg', 'jpeg'):`

Issue #150, opened 2016-12-10
Timezone error in media endpoint


Leonidaz0r:
Some requests to the media endpoint seem to generate errors. From the logfile:

Error on request:
Traceback (most recent call last):
  File "/home/amiv/amivapi/pyenv/lib/python3.5/site-packages/werkzeug/serving.py", line 194, in run_wsgi
    execute(self.server.app)Error on request:
Traceback (most recent call last):
  File "/home/amiv/amivapi/pyenv/lib/python3.5/site-packages/werkzeug/serving.py", line 194, in run_wsgi
    execute(self.server.app)
  File "/home/amiv/amivapi/pyenv/lib/python3.5/site-packages/werkzeug/serving.py", line 182, in execute
    application_iter = app(environ, start_response)
  File "/home/amiv/amivapi/pyenv/src/eve/eve/flaskapp.py", line 1014, in __call__
    return super(Eve, self).__call__(environ, start_response)
  File "/home/amiv/amivapi/pyenv/lib/python3.5/site-packages/flask/app.py", line 1836, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/amiv/amivapi/pyenv/lib/python3.5/site-packages/flask/app.py", line 1820, in wsgi_app
    response = self.make_response(self.handle_exception(e))
  File "/home/amiv/amivapi/pyenv/lib/python3.5/site-packages/flask/app.py", line 1403, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/amiv/amivapi/pyenv/lib/python3.5/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/home/amiv/amivapi/pyenv/lib/python3.5/site-packages/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/amiv/amivapi/pyenv/lib/python3.5/site-packages/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/amiv/amivapi/pyenv/lib/python3.5/site-packages/flask/app.py", line 1381, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/amiv/amivapi/pyenv/lib/python3.5/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/home/amiv/amivapi/pyenv/lib/python3.5/site-packages/flask/app.py", line 1475, in full_dispatch_request
[...]

PR #155, opened 2016-12-12
Fix for #150

Leonidaz0r:
Eve expects the file updated time to have UTC tzdata. See issue #150

### DIFF
--- a/amivapi/media.py
+++ b/amivapi/media.py
@@ -10,7 +10,7 @@ Saves Uploaded media to a folder specified in config['STORAGE_FOLDER']
 
 """
 
-from bson import tz_util
+import pytz
 from os import path, remove, makedirs, urandom
 import errno
 from contextlib import contextmanager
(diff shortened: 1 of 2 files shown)


amivapi/media.py
```diff
@@ -10,6 +10,7 @@
 
 """
 
+from bson import tz_util
```
NotSpecial: Why are we importing the tz from bson? This seems a odd place to take timezone information from.

NotSpecial: Also could you make add a test for this problem? Or I'll try to add it later if I have time.

Leonidaz0r: I basically used the tz from bson, because eve also did. Interesting question. I guess eve uses it, because they don't want to depend on pytz. However that is bad anyway, as eve depends on bson in that way, even if the mongo data layer is not used.
I am not sure whether there is any difference apart from what dependencies to libraries are generated.

Issue #158, opened 2016-12-13
JSON schema vs Cerberus schema


Leonidaz0r:
Currently we use cerberus schemas for event's additional_fields. There was a discussion, whether we should switch to JSON schemas, as those are more widely known outside the python community.

Thoughts?

hermannsblum:
As there is a [jsonschema validation for python](https://pypi.python.org/pypi/jsonschema), but no cerberus schema validation for js, that would make client-side validation easier

NotSpecial:
Aren't there some javascript libraries around to create jsonschema? And to create forms form jsonschema?

In my opinion something like this would be the main argument, because if something like this exists we would make life for the frontend team much easier.

hermannsblum:
Most of the translation to json schema is done in #145 .
However, the `additional_fields` Are still cerberus. This should be done with a PR to our version of eve-swagger

PR #159, opened 2016-12-13
Improvements of signups

Leonidaz0r:
Restructured the tests, generally tidied up code of event module.
There is now a field `selection_strategy` in events and `accepted` in eventsignups. These are used for different methods to give spots to people. FCFS and manual setting are supported. Also emails are sent, when people are automatically admitted to an event. The emails contain a link to give up the spot and let other people get promoted.

### DIFF
--- a/amivapi/events/model.py
+++ b/amivapi/events/model.py
@@ -181,7 +181,7 @@ eventdomain = {
             },
             'additional_fields': {
                 'nullable': True,
-                'type': 'json_schema',
+                'type': 'cerberus_schema',
                 'only_if_not_null': 'spots'
             },
             'allow_email_signup': {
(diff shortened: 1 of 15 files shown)


NotSpecial:
Why have you switched the additional fields back to cerberus schema?

It seems to me the concensus in #158 was thay jsonschema is better suited here.

amivapi/events/model.py
```diff
@@ -181,7 +182,7 @@
             },
             'additional_fields': {
                 'nullable': True,
-                'type': 'json_schema',
+                'type': 'cerberus_schema',
```
NotSpecial: I think this change is a step back.

Leonidaz0r: no it is just a name change. Look here https://github.com/amiv-eth/amivapi/pull/159/commits/47ec1aa0cda86cf0c93bc104f396948d3823e622#diff-95bf83f7b01c7278d2a3d182af15cc19R145
The name json_schema was misleading, because it IS a cerberus schema we are parsing.

NotSpecial: Ah, i see! We should definitely change this, but that has nothing to do witu this pull request then.

PR #160, opened 2016-12-13
Exclude ldap, command line tool and tests from coverage report

Leonidaz0r:
ldap and command line tool will probably never get coverage. Enforcing coverage of the test does not make sense at many locations as well.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,5 @@
+[run]
+omit =
+     amivapi/ldap.py
+     amivapi/cli.py
+     amivapi/tests/*


fubu:
I agree with Alex that the coverage of the tests itself might be interesting (at work we are doing it for the same reason Alex mentioned, and it already led to test refactorings a few times).

In general, what exactly is the problem with the coverage? Do you just want to "increase" the coverage number, or is there another reason?

Leonidaz0r:
Main reason was, that I looked at what we miss in the tests and most missed lines are exceptions which should really not be hit like the database not being empty or a fixture being broken. I guess we could write tests testing the tests, if we really want to. Other option would be to just exclude those lines.

fubu:
"Tests testing the tests" sounds... weird :) At work we just mark such lines with a `# pragma: no cover` comment, which excludes the whole branch (e.g. `except` or `else`) from coverage. But having this coverage for tests gives you a tool to detect possibly unused helper code in your utils.

BTW, the other reason why I would keep the ldap and CLI module under coverage: It shows you that there is a whole bunch of production code not being tested for correctness (and I do think it's possible to write at least some tests for both modules!). On the other hand, the single lines you mentioned which just make sure tests don't leave traces? Not relevant in production, so excluding these lines/branches is fine.

NotSpecial:
To dig this up again: I agree with fubu and therefore would not merge this. 

I'll get us some CLI tests and for the rests I'd go with fubus suggestion

PR #163, opened 2017-03-06
Studydoc patches

NotSpecial:


Leonidaz0r:
Maybe a wildcard delete for sessions might be interesting. However we can still do that on the client side...

Issue #91, opened 2015-11-14
Provide API Schema in jsonschema format


NotSpecial:
The javascript acces library (handling most api request in the amiv web apps) should contain a input validator in the future. For this reason it would be nice if we could use an automatically generated schema.

the docs/json.spec endpoint already does this but returns our cerberus schema.

It would be nice if we could return it in the jsonschema standard because there are javascript solutions to use this directly for validation.
(Using the cerberus file we would have to parse it all ourselves)

I think it should not be to complicated to generate the jsonschema from our cerberusschema.

http://json-schema.org/

hermannsblum:
To my knowledge the cerberus-schema actually is a json-schema. So the parsing should not be a problem.
Is it because of the special data-types?

hermannsblum:
Fixed as swagger also uses jsonschema #145

Leonidaz0r:
not fixed yet

hermannsblum:
This is (now really merged) done in #145.
There is still the issue of `additional_fields`, see #158. However, I am not sure whether we want 2 issues open for the same problem. Can we close this?

NotSpecial:
Done now! (Or in a second when travis finishes)

PR #165, opened 2017-03-07
use jsonschema in additonal_fields of events

### LINKED ISSUE #158: JSON schema vs Cerberus schema

hermannsblum:
This PR fixes #158 and #91, aswell as a minor bug in some event-validator.

*Debatable*: json-schema are more complex than cerberus-schema. I now require the user to specify a full json-schema for additional_fields, while ensuring in the validator that the following fields are always set accordingly:

    {
           "$schema": "http://json-schema.org/draft-04/schema#",
           "type": "object",
           "additionalProperties": False
    }

This is necessary to ensure that additional_fields specifies a dict and not a list (which makes it easier for an UI to create an understandable, dynamic form for these fields), aswell as to ensure the correct usage of json-schema to our purposes. One could also overwrite this api-side, but it would be instransparent to let somebody submit a non-complete json-schema and add or overwrite values.

### DIFF
--- a/amivapi/events/validation.py
+++ b/amivapi/events/validation.py
@@ -238,7 +238,7 @@ class EventValidator(object):
             self._original_document is not None and
             self._original_document.get(only_if_not_null) is not None)
 
-        if not(doc.get(only_if_not_null) is not None or exists_in_original):
+        if doc.get(only_if_not_null) is None and not exists_in_original:
             self._error(field, "May only be specified if %s is not null"
                         % only_if_not_null)
 
--- a/amivapi/events/validation.py
+++ b/amivapi/events/validation.py
@@ -236,7 +236,12 @@ class EventValidator(object):
             field (string): name of the validated field
             value: Value of the validated field
         """
-        if self.document.get(only_if_not_null) is None:
+        doc = self.document
+        if self._original_document is not None:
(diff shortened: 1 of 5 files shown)


NotSpecial:
I simplified the only if not null validator and added some tests to check the patching explicitly.

amivapi/events/validation.py
```diff
@@ -236,17 +234,16 @@ def _validate_only_if_not_null(self, only_if_not_null,
...
-            # This is a patch. The patch-document will not contain all
-            # information
-            doc = deepcopy(self._original_document)
-            doc.update(self.document)
-        if doc.get(only_if_not_null) is None:
+        exists_in_original = (  # Check original document in case of patches
+            self._original_document is not None and
+            self._original_document.get(only_if_not_null) is not None)
+
+        if not(doc.get(only_if_not_null) is not None or exists_in_original):
```
Leonidaz0r: I think this would be simpler:
```
if doc.get(only_if_not_null) is None and not exists_in_original:
```

NotSpecial: Sounds good

PR #169, opened 2017-03-19
Apikeys as resource

NotSpecial:
I was starting to write tests for the CLI when I thought: Do API Keys really need to be handled with CLI?

I think they would work better as an API resource.

Benefits:
- They could be managed more easily with admintools
- With database backups API keys will also be safed
- We don't need an additional config file for the keys.

Since the functionality provided is the same I can't see any downsides.

Instead of creating an issue I thought I'd go ahead an just implement it. What are your opinions?

### DIFF
--- a/amivapi/auth/apikeys.py
+++ b/amivapi/auth/apikeys.py
@@ -3,38 +3,107 @@
 # license: AGPLv3, see LICENSE for details. In addition we strongly encourage
 #          you to buy us beer if we meet and you like the software.
 
-"""Hook to allow authorization with apikeys."""
+"""Authorization with apikeys.
+
+Provides the apikey resource and hooks to handle authorization with a key.
+
+API keys should only be created or modified by admins.
+"""
+from os import urandom
+from base64 import b64encode
+from datetime import datetime as dt
 
 from flask import abort, current_app, g
 
-from ruamel import yaml
+from amivapi.utils import register_domain
+
+from .auth import AmivTokenAuth
 
 
 def authorize_apikeys(resource):
-    """Check if user is an apikey, and if it is, do authorization."""
-    permissions = [
-        data['permissions'] for data in current_app.config['APIKEYS'].values()
-        if data['token'] == g.get('current_token')]
-    if permissions:
-        # We can just take first (and only) element since token is unique
-        permission = permissions[0].get(resource)
+    """Check if user is an apikey, and if it is, do authorization.
+
+    Also update 'updated' timestamp everytime a key is accessed
+    """
+    apikeys = current_app.data.driver.db['apikeys']
+    apikey = apikeys.find_one({'token': g.get('current_token')},
+                              {'permissions': 1})
+
+    if apikey:
+        # Get permission for resource if they exist
+        permission = apikey['permissions'].get(resource)
+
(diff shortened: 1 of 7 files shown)


amivapi/auth/apikeys.py
```diff
@@ -3,38 +3,107 @@
...
+            'token': {
+                'type': 'string',
+                'readonly': True,
+                'description': 'The token that can be used for auth. '
+                'Will be randomly generated for every new key.'
+            },
+            'permissions': {
+                'type': 'dict',
+                'propertyschema': {'type': 'string',
+                                   'api_resources': True},
```
Leonidaz0r: This validator is removed from groups/validation.py further down. How does this even work?

amivapi/auth/apikeys.py
```diff
@@ -3,38 +3,107 @@
...
+            abort(403, "The API key exists but does not grant the required "
+                       "permissions.")
+
+
+class ApikeyAuth(AmivTokenAuth):
+    """Do not let (non-admin) users view API keys at all.
+
+    We don't want to filter the output , we want to abort completely."""
+    def create_user_lookup_filter(self, user):
+        abort(403)
```
Leonidaz0r: This is a weird way to manage the authorization... What is the intention?
I would think this should just be done via the normal group based authorization.

NotSpecial: I don't quite understand your question. What is weird about it? This is unchanged from the previous implementation of api keys. 

And yes, the mechanism is practically identical to groups, i.e. Api keys grant permissions in the same ways groups do. Nevertheless groups are only for real users and api keys exists to give a set of permissions to applications. 

What part exactly do you think should be done in groups? And how?

amivapi/auth/apikeys.py
NotSpecial: It is added to base validator in utils.py since it's now a function that is used in several modules

NotSpecial:
Regarding tests: I have added [a lot](https://github.com/amiv-eth/amivapi/blob/d378572e372ccfa40536edfdb452fffbe39dac93/amivapi/tests/auth/test_apikeys.py) of them, maybe you missed this file? Or what tests do you think are missing?

amivapi/auth/apikeys.py
NotSpecial: Or do you maybe talk about the abort(403)? This exists mainly to provide a better error message and to distinguish it from using a bad key. Maybe I should add a comment explaining this.

Leonidaz0r:
Hmm it seems I reviewed different code. My version looked a lot shorter. Let me check again. Probably I messed up and viewed a single commit.

amivapi/auth/apikeys.py
Leonidaz0r: Ah now I understand how it works. I thought this was a way to prevent anyone without the root token from POST/GET to /apikeys. So nevermind my comment. I slightly modified the docstring, as Tar-Tarus also seems to have had trouble with this.

amivapi/auth/apikeys.py
Leonidaz0r: You are correct.

NotSpecial:
Great, thanks for the updated comments, it's easier to understand now. Merging :)

PR #171, opened 2017-03-20
Make flake8 enforce import order and no builtin shadowing

Leonidaz0r:
I added plugins to flake8 for further rule checking. This should be merged after the cleanup branch, which fixes all errors discovered by these settings.

For me it also works with my editor, if I install the two plugins system wide and add these lines(same as tox.ini) to my global flake8 config:

```
ignore = I201
application-import-names = amivapi
import-order-style = google
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@
 
 """Install amivapi. Provide the amivapi command."""
 
-from setuptools import setup, find_packages
+from setuptools import find_packages, setup
 
 with open('LICENSE') as f:
     LICENSE = f.read()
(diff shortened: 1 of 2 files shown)


Leonidaz0r:
I added comments.

NotSpecial:
In other news: isn't it kind of inconvenient now that we have already merged "cleaned" while this is just an addendum to "cleanup"?^^

Leonidaz0r:
I created a new pull request against the master branch https://github.com/amiv-eth/amivapi/pull/173

PR #173, opened 2017-03-22
Flake8 plugins rebased

Leonidaz0r:
Rebased version of Pull request https://github.com/amiv-eth/amivapi/pull/171


Leonidaz0r:
We decided we don't want this.

PR #174, opened 2017-04-16
LDAP improvements the third

NotSpecial:
First of all, the structure of ldap.py is now simplified. The class is dropped
and replaced by a few module level functions which are easier to understand.

Next the functions themselves have been rearranged to be easier to test.
Some premature optimization has been removed: The updating of all users
now uses the same function as the one for one user only, since there were not
really noticable performance differences. Again, this simplifies testing.

Speaking of testing: Using mock, all ldap functions are now tested.
There still are some integration tests to verify everything actually works with the
eth ldap.

Finally a few changes in bootstrap, cli and sessions to use the new functions.

After all I am quite hapy with this version now. A few points are still up for improvement and I have added them to ldap.py, but over all the code is much simpler and better tested now, which is nice.

I think I messed up the history a little while rebasing and some commits show up twice and I honestly have no idea how that happened. If anyone knows how this can be fixed please help.

### DIFF
--- a/amivapi/auth/sessions.py
+++ b/amivapi/auth/sessions.py
@@ -15,9 +15,9 @@ from eve.methods.patch import patch_internal
 from eve.utils import config, debug_error_message
 from flask import abort, current_app as app
 
+from amivapi import ldap
 from amivapi.auth import AmivTokenAuth
 from amivapi.cron import periodic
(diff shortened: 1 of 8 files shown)


NotSpecial:
I'll open another pull request without destroyed history.

PR #175, opened 2017-04-16
Ldap improvements

NotSpecial:
First of all, the structure of ldap.py is now simplified. The class is dropped
and replaced by a few module level functions which are easier to understand.

Next the functions themselves have been rearranged to be easier to test.
Some premature optimization has been removed: The updating of all users
now uses the same function as the one for one user only, since there were not
really noticable performance differences. Again, this simplifies testing.

Speaking of testing: Using mock, all ldap functions are now tested.
There still are some integration tests to verify everything actually works with the
eth ldap.

Finally a few changes in bootstrap, cli and sessions to use the new functions.

After all I am quite happy with this version now. A few points are still up for improvement and I have added them to ldap.py, but over all the code is much simpler and better tested now, which is nice.

### DIFF
--- a/amivapi/tests/test_ldap.py
+++ b/amivapi/tests/test_ldap.py
@@ -29,37 +29,40 @@ class LdapTest(WebTestNoAuth):
 
     def test_init_app(self):
         """Test that init app uses correct credentials stored connector."""
-        self.app.config['LDAP_USER'] = 'test'
-        self.app.config['LDAP_PASS'] = 'T3ST'
+        ldap_user = 'test'
+        ldap_pass = 'T3ST'
+        initialized_ldap = 'I totally am an ldap instance.'
+
+        self.app.config['LDAP_USER'] = ldap_user
+        self.app.config['LDAP_PASS'] = ldap_pass
         to_patch = 'amivapi.ldap.AuthenticatedLdap'
-        with patch(to_patch, return_value='initialized') as init:
+
+        with patch(to_patch, return_value=initialized_ldap) as init:
             ldap.init_app(self.app)
 
-            init.assert_called_with('test', 'T3ST')
-            self.assertEqual(self.app.config['ldap_connector'], 'initialized')
+            init.assert_called_with(ldap_user, ldap_pass)
+            self.assertEqual(self.app.config['ldap_connector'],
+                             initialized_ldap)
 
     def test_ldap_auth(self):
         """Test that ldap can authenticate a user."""
+        nethz = "Pablo"
+        login_data = {'username': nethz, 'password': 'p4bl0'}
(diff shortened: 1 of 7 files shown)


amivapi/tests/test_ldap.py
```diff
@@ -0,0 +1,275 @@
...
+
+    def test_init_app(self):
+        """Test that init app uses correct credentials stored connector."""
+        self.app.config['LDAP_USER'] = 'test'
+        self.app.config['LDAP_PASS'] = 'T3ST'
+        to_patch = 'amivapi.ldap.AuthenticatedLdap'
+        with patch(to_patch, return_value='initialized') as init:
+            ldap.init_app(self.app)
+
+            init.assert_called_with('test', 'T3ST')
```
fubu: magic strings

amivapi/tests/test_ldap.py
```diff
@@ -0,0 +1,275 @@
...
+            self.assertEqual(self.app.config['ldap_connector'], 'initialized')
```
fubu: magic string

amivapi/tests/test_ldap.py
```diff
@@ -0,0 +1,275 @@
...
+            'sn': ['Ablo'],
+            'swissEduPersonGender': '0',
+            'ou': ['VSETH Mitglied', 'D-ITET',
+                   'Informationstechnologie und Elektrotechnik'],
+            'some_random_field': 'abc',
+        }
+        data.update(kwargs)
+        return data
+
+    def fake_filtered_data(self):
```
fubu: This method returns static data - why not using a variable instead?

requirements.txt
```diff
@@ -8,5 +8,7 @@ click==6.6
 pytz==2016.7
 ruamel.yaml==0.12.14
 freezegun==0.3.8
+# Mock is only available in the standart lib from 3.3 on
```
fubu: standar**d** :)

NotSpecial: Oops ^^

NotSpecial:
Hi @fubu, thanks for your review!

Could you elaborate on your 'magic string' comments, please? I don't quite understand what you mean.

fubu:
Magic strings (also numbers) are "constants" with a specific meaning in the application code which is not obvious from first looking at it; [some reference](http://deviq.com/magic-strings/). In the lines I commented on you repeat such magic strings like passwords or usernames, which you defined before - it's better practice to store them in a variable and reuse that later, as you did in most other tests (`expected` et al).

NotSpecial:
Ah, I see. Yes, in the short tests I skipped using a variable. But I see why its better to use a variable. I will adjust it

PR #189, opened 2017-10-29
Remove trailing = in tokens.

Leonidaz0r:
They are unsafe in URLs and add no randomness


Leonidaz0r:
wait this is already in the master...

Issue #176, opened 2017-06-03
Validators may crash if they use a value that has not been type checked


cburchert:
E.g. posting the date "2017-00-00T00:00:00Z" for time_start in an event will give an exception and not return a proper error message.

Validators are run in a bad order. Cerberus iterates over the fields and runs all validators for the current field. This results in the following order of validators:

time_end: _validate_type
time_end: _validate_later_than
time_start: _validate_type

A wrong type for time_start can crash the _validate_later_than function. A badly formatted date string will result in a string object instead of a datetime object(eves policy for type errors is to keep a string) and should raise a type error. This is hidden by the validator order.

The same problem appears for all other validators using the value of a different field than the one containing the rule.

Leonidaz0r:
Analysis:
The problem only appears in certain fields. The bad date must be a target of the later_than validator.
Normally for any field first the type is checked, then other validators are run. However the later_than validator uses another field before the field itself has been validated, resulting in a crash in the validator if the type is wrong.

Leonidaz0r:
Test added in the new validator_fail branch: https://github.com/amiv-eth/amivapi/tree/validator_fail

PR #193, opened 2017-12-01
Added test for issue #176

cburchert:


temparus:
Are you still working on this, @cburchert?

cburchert:
I haven't looked into it for some time. I remember abandoning it because I
found no good way to solve this.

NotSpecial:
I have now fixed the issue by explicitly parsing the datetime string. If it fails, be can ignore it, as the normal validator will complain.

PR #196, opened 2017-12-02
auth hooks can work with embedded relations

### LINKED ISSUE #195: API crashes if user and group is embedded in GET to groupmemberships
The auth functions assume that item['user'] in returned objects from the database are object ids. However with embedding this is not the case.
One example is here:
```
 def has_item_write_permission(self, user_id, item):
        """The group moderator and the member can change an enrollment."""
        if user_id == str(item['user']):
            # Own membership can be modified
            return True
        else:
            # Check if moderator
            # Note: Group must exist, otherwise membership would not exist
            #   Furthermore user_id can't be None so if there is no moderator
            #   we will correctly return False
            collection = current_app.data.driver.db['groups']
            group = collection.find_one({'_id': item['group']},
                                        {'moderator': 1})
            return user_id == str(group.get('moderator'))
```

item['user'] and item['group'] can both be objects if they are embedded via embedded clause in the query. Then this function crashes.

Similar constructs are also in other functions.

Leonidaz0r:
Fixes #195

### DIFF
--- a/amivapi/utils.py
+++ b/amivapi/utils.py
@@ -5,6 +5,7 @@
 """Utilities."""
 
 
+from bson import ObjectId
 from contextlib import contextmanager
 from copy import deepcopy
 from email.mime.text import MIMEText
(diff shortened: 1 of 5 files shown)


amivapi/utils.py
```diff
@@ -35,6 +36,23 @@ def admin_permissions():
...
+    Args:
+        item: Either an object from the database as a dict or an object id as
+            str or objectid.
+
+    Returns:
+        ObjectId with the user ID
+    """
+    if isinstance(item, dict):
+        return ObjectId(item['_id'])
+    return ObjectId(item)
```
NotSpecial: I think a more "pythonic" version would be to use `try-catch`

PR #202, opened 2017-12-03
Media: Remove custom filestorage class

NotSpecial:
Since we are going to use docker, it's just simpler to store everything in the database.


Leonidaz0r:
When downloading a file, the filename is a random string. It would be better if the path is correct. Can we fix this somehow?

NotSpecial:
I do not think eve supports this, it uses the gridfs key for the filename. A html client could set the name with `<a href='...' download=filename>` though, I think. The filename is only included in the meta data, isn't it?

Leonidaz0r:
ok. Maybe something Eve could fix. However the client side solution should be fine for us, as some intelligent client needs to get the URL first anyway.

PR #207, opened 2017-12-06
Use SSL to send mails

Leonidaz0r:


Leonidaz0r:
That is exactly what I thought, but when I tried it, I figured the server doesn't want it. Try in a python terminal:
```
import smtplib
s = smtplib.SMTP("smtp.ee.ethz.ch")
s.starttls()
s.login("<nethz>", "<Password>")
```

The server rejects it stating it does not support login.

Moschn:
Try this:

``` python
import smtplib
s = smtplib.SMTP(host="smtp.ee.ethz.ch", port=587)
s.starttls()
s.ehlo()
s.login("<nethz>", "<Password>")
```

amivapi/cli.py
```diff
@@ -107,6 +107,15 @@ def no_ldap_prompts(ctx, param, value):
 @option("--smtp", "SMTP_SERVER", default="localhost",
         prompt="STMP server for outgoing mails",
         help="SMTP server.")
+@option("--smtp", "SMTP_PORT", default="0",
```
Moschn: port should probably default to 25 or 587

amivapi/settings.py
```diff
@@ -45,8 +45,10 @@
 
 DEFAULT_CONFIG_FILENAME = 'config.yaml'
 
-# Timeout for connections to the SMTP server
-SMTP_TIMEOUT = 3
+# SMTP server defaults
+SMTP_HOST = 'localhost'
+SMTP_PORT = 0
```
Moschn: port should probably default to 25 or 587

amivapi/utils.py
```diff
@@ -76,20 +76,25 @@ def mail(sender, to, subject, text):
...
-            s.quit()
+            with smtplib.SMTP(config.SMTP_SERVER,
+                              port=config.SMTP_PORT,
+                              timeout=config.SMTP_TIMEOUT) as smtp:
+                status_code, _ = smtp.starttls()
+                if status_code != 220:
+                    app.logger.error("Failed to create secure "
+                                     "SMTP connection!")
+                    return
+
```
Moschn: Add `smtp.ehlo()`

see [documentation](https://docs.python.org/3/library/smtplib.html#smtplib.SMTP.starttls)

NotSpecial:
@Moschn Login will send the ehlo() automatically. (see docs) 

@leonidaz0r have you maybe forgotten then port? The ee server only responds on port 587, which is not the default port chosen by pythons SMTP. 

It definitely worked for me, I used the api to send you those mails yesterday ^^

Moschn:
@NotSpecial Ah ok. Then no ehlo is needed

NotSpecial:
It kind of is, I just noticed: if we don't need a login, we have to send the ehlo manually

amivapi/settings.py
NotSpecial: Since we are using starttls, we should probably use 587, I agree

Issue #180, opened 2017-10-09
Remindermails


Leonidaz0r:
We could add an option to automatically send reminder mails for events.

PR #209, opened 2017-12-23
Add Reminder Email for events

Tar-Tarus:
#180 
Add the creation and sending of a reminder email for events. The reminder will be sent to all the users signed up for the event.

### DIFF
--- a/amivapi/cron.py
+++ b/amivapi/cron.py
@@ -97,9 +97,12 @@ def periodic(period, *args):
     return wrap
 
 
-def schedule_task(time, func, *args):
+def schedule_task(time, func, *args, func_alternative_name = ""):
     """ Schedule a task at some point in the future. """
-    func_s = func_str(func)
+    if alternative_name is not None:
+        func_s = func_alternative_name
+    else:
+        func_s = func_str(func)
 
     if func_s not in schedulable_functions:
         raise NotSchedulable("%s is not schedulable. Did you forget the "
@@ -112,9 +115,12 @@ def schedule_task(time, func, *args):
     })
 
 
-def update_scheduled_task(time, func, *args):
+def update_scheduled_task(time, func, *args, func_alternative_name = ""):
     """ Update a scheduled task that was previously registered. """
-    func_s = func_str(func)
+    if func_alternative_name is not "":
+        func_s = func_alternative_name
+    else:
+        func_s = func_str(func)
 
     if func_s not in schedulable_functions:
         raise NotSchedulable("%s is not schedulable. Did you forget the "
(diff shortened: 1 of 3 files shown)


amivapi/cron.py
```diff
@@ -99,7 +99,11 @@ def wrapped():
 
 def schedule_task(time, func, *args):
     """ Schedule a task at some point in the future. """
-    func_s = func_str(func)
+    if func_str(func) is "remindermail":
+        item = pickle.load(args)
+        func_s = "remindermail"+str(item['_id'])
+    else:
+        func_s = func_str(func)
```
NotSpecial: This seems super hacky to me. Is this special case for reminder mails really necessary?

amivapi/cron.py
```diff
@@ -114,7 +118,11 @@ def schedule_task(time, func, *args):
 
 def update_scheduled_task(time, func, *args):
     """ Update a scheduled task that was previously registered. """
-    func_s = func_str(func)
+    if func_str(func) is "remindermail":
+        item = pickle.load(args)
+        func_s = "remindermail_"+str(item['_id'])
```
NotSpecial: I think this will break the update. the `update_one` below identifies the function by name, so if the name is changed, the original function cannot be updated anymore.

amivapi/events/emails.py
```diff
@@ -59,6 +61,91 @@ def send_confirmmail_to_unregistered_users_bulk(items):
...
+        time_start_event = datetime.strptime(item['time_start'],
+                                             '%Y-%m-%dT%H:%M:%SZ')
+        datetime_reminder = time_start_event - \
+            timedelta(
+                  days=int(current_app.config['REMINDER_EMAIL_DAYS2EVENT']))
+        update_scheduled_task(datetime_reminder,
+                              remindermail,
+                              item)
+
+
```
NotSpecial: We have already discussed this in the chat shortly -- might it not be easier to add a periodic task sending reminder mails for all upcoming events?

amivapi/events/emails.py
```diff
@@ -59,6 +61,91 @@ def send_confirmmail_to_unregistered_users_bulk(items):
...
+        date_time_event = " NaN "
+
+    fields = {
+        'location': location,
+        'datetime': date_time_event,
+        'title': title
+    }
+
+    # Populate content text with fetched infos
+    email_content = current_app.config['REMINDER_EMAIL_TEXT'] % fields
```
NotSpecial: As fair as I can see, `REMINDER_EMAIL_TEXT` is not in the current app settings. I think we should provide a default, similar to the signup confirmation

temparus:
We have recently moved to templated email messages using jinja2 and send multipart email messages with an HTML and a plaintext version.

Due to this change, this pull request must probably be reworked as it is quite old now.

PR #217, opened 2017-12-31
Add Cache-Control: must-revalidate to default settings (resolves #214)

### LINKED ISSUE #214: Add Cache-Control header to API response
It happend a couple of times that the browser served the API request from the browser cache when reloading the page, which is absolute nosense.

To avoid this, the following header should be added to every response:
```
Cache-Control: no-store, must-revalidate
```

temparus:


amivapi/settings.py
```diff
@@ -35,6 +35,7 @@
 X_DOMAINS = '*'
 X_HEADERS = ['Authorization', 'If-Match', 'Content-Type']
```
hermannsblum: I think it makes sense to add `Cache-Control` here to allow users to set their own cache settings.

hermannsblum:
Hmm, for some reason code-comments are not referrenced here^^
To repeat, as this PR adds a default setting for the `Cache-Control` header, I think it should also allow client-defined cache settings.

amivapi/settings.py
temparus: Why should this be allowed? 
I don't see why the client should specify this header in a request.

NotSpecial: I agree with @hermannsblum -- as far as I understand, the API will send out a cache-control header to the client, but if the client were to respond using this (provided) header, the browser would actually prevent the client from making the request (since `Cache-Control` is not one of the allowed cross domain headers)

PR #218, opened 2017-12-31
Fix documentation of event.additional_fields (resolves #212)

### LINKED ISSUE #212: Wrong documentation for entity event
For the additional_fields of event the documentation / description is wrong.
It should be the following:

```
must be provided in form of a JSON-Schema. Youcan add here fields you want to know from people signing up going further than their email-address. The JSON-Schema will always require these fields: {"$schema": "http://json-schema.org/draft-04/schema#","type": "object","additionalProperties": false}
```

temparus:


temparus:
@hermannsblum @NotSpecial please review

Issue #219, opened 2017-12-31
API crashes if Authentication header is wrong formatted


temparus:
The API crashes if the request header contains `Authentication: Token` without providing an actual token.

Here's the traceback when the described situation occurs:
```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/eve/flaskapp.py", line 1033, in __call__
    return super(Eve, self).__call__(environ, start_response)
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1836, in __call__
    return self.wsgi_app(environ, start_response)
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1820, in wsgi_app
    response = self.make_response(self.handle_exception(e))
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1403, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1381, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/local/lib/python2.7/dist-packages/eve/endpoints.py", line 54, in collections_endpoint
    response = get(resource, lookup)
  File "/usr/local/lib/python2.7/dist-packages/eve/methods/common.py", line 284, in rate_limited
    return f(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/eve/auth.py", line 79, in decorated
    return f(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/eve/methods/common.py", line 1032, in decorated
    getattr(app, event_name)(*gh_params)
  File "/usr/local/lib/python2.7/dist-packages/events/events.py", line 77, in __call__
    f(*a, **kw)
  File "./amivapi/auth/auth.py", line 221, in authenticate
[...]

PR #220, opened 2017-12-31
Fix issue #219

temparus:
Prevent the API from crashing when an empty token is provided for authentication.

### DIFF
--- a/amivapi/auth/auth.py
+++ b/amivapi/auth/auth.py
@@ -217,7 +217,7 @@ def authenticate(*args):
     # Code copied from Eve's TokenAuth - parse different header formats
     if not token and request.headers.get('Authorization'):
         token = request.headers.get('Authorization').strip()
-        if token.lower().startswith(('token', 'bearer')):
+        if token.lower().startswith(('token', 'bearer')) and ' ' in token:
             token = token.split(' ')[1]
     # End of code copied from Eve
 
--- a/amivapi/auth/auth.py
+++ b/amivapi/auth/auth.py
@@ -198,8 +198,7 @@ def authenticate(*args):
     Recognized tokens in following formats:
     - Basic auth with token as user and no password
     - "Authorization: <token>"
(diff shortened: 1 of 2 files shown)


temparus:
@cburchert @NotSpecial please review

NotSpecial:
#Lernphase

Our coding day is next saturday -- I'll have a look then.

NotSpecial:
Ok wait, it's just one line -- I have time for that^^

NotSpecial:
Looks alright to me, but can you add a test that demonstrates the crashing?

NotSpecial:
One line was too long and flake8 failed. I fixed this and will merge as soon as the checks pass.

NotSpecial:
Ok, while fixing flake8 errors, I decided that this is all much to complicated and simplified the token parsing.(see description in 7f9362f)

@cburchert , could you have a quick look before the squash and merge?

amivapi/auth/auth.py
```diff
@@ -211,15 +210,13 @@ def authenticate(*args):
...
-
-    # Code copied from Eve's TokenAuth - parse different header formats
-    if not token and request.headers.get('Authorization'):
-        token = request.headers.get('Authorization').strip()
-        if token.lower().startswith(('token', 'bearer')):
-            token = token.split(' ')[1]
-    # End of code copied from Eve
+    # Get token: First try basicauth username, else just auth header
+    token = (getattr(request.authorization, 'username', '') or
+             request.headers.get('Authorization', '').strip())
```
cburchert: Somehow this line worries me. This value is basically user controlled and given to the database in line 226.
Are we sure only a simple string will come out of here? Otherwise there might be a way to at least bypass authentication.

NotSpecial: As far as I know, a http header can *only* be a string

cburchert: I looked into it a bit and it seems to be the case.

Issue #215, opened 2017-12-31
Add field authentication_method to user entity


temparus:
We should add an additional required field `authentication_method` with the values `"ldap"` and `"password"` to the entity user.

For transparency purpose, the documentation should be changed as follows:
The value of `password` is only used when `authentication_method` is set to `password`.

NotSpecial:
Why is this required?

If a password is set, the api tries to use it for authentication. Otherwise it uses LDAP.

What do we get from adding this field?

hermannsblum:
As from the chat: The basic problem is that there needs to be a "change your password" option for those users that do not authenticate with LDAP and you need to know which users should be displayed this option and which not.

NotSpecial:
Ok, so the main issue is that a frontend does not know whether a user has a password set or not?

In this case I see why we would need this and I would go with the original proposal, since as you said it's the cleanest interface.

NotSpecial:
On the other hand: when using an additional field, the frontend still can only tell if the user *wants* to use a password -- it can't tell if the user already has set one.

Maybe we need to provide feedback some other way. E.g. using a boolean field that is true if a password is set and false if not? This could be done easily in a hook, I think.

NotSpecial:
Other idea: the hook currently removing the password field could maybe only remove the password field for other users? And for the users own password, it could replace it with some '*******' or similar.

This way other users don't get any info about how you are authenticating and you can tell if you have set a password (use pw auth) or not (use LDAP)

temparus:
I think that your last idea would be the easiest one to implement and the most intuitive one too.

@Leonidaz0r @hermannsblum What do you think about that?

hermannsblum:
As already mentioned, I basically also see these two options.

However, setting the return value of the password field to some defined arbitrary value is really the least intuitive way, as it is really not standard REST that a field should have very different contents for different methods (in GET the `password` field now shows whether or not the user has set a password, in PATCH and POST it however should never be set, except for when the user wants to set the password to the defined value). This can easily lead to confusion in client implementations and may result in accidentially setting a password (e.g. by GETting an opject, doing some manipulations on it and PATCHing it again). The better option in my opinion is therefore really to hide the password field alltogether and add the additional boolean field. I think the form as described in the original issue description would be the best understandable.

cburchert:
[...]

PR #223, opened 2018-01-27
GET to /users now projects a password_set field

cburchert:
Discussion in issue #215

### DIFF
--- a/amivapi/users/security.py
+++ b/amivapi/users/security.py
@@ -110,6 +110,9 @@ def hide_fields(response):
 def project_password_status(response):
     """Add a boolean field password_state to the response.
 
+    This function must be applied before hide_fields, as it uses the password
+    field.
+
     Args:
         response: Response object of the request
     """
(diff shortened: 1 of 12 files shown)


amivapi/users/security.py
```diff
@@ -105,6 +105,22 @@ def hide_fields(response):
                     item.pop(key)
 
 
+# Project password status
+
+def project_password_status(response):
```
NotSpecial: This function can only work if it is applied *before* hide_fields, otherwise there will be no password field. This should be mentioned somewhere.

amivapi/users/security.py
```diff
@@ -105,6 +105,22 @@ def hide_fields(response):
...
+
+    Args:
+        response: Response object of the request
+    """
+    # Compatibility with both item and resource hook
+    items = response.get('_items', [response])
+
+    for item in items:
+        item['password_set'] = item.get('password') is not None
+        print(item)
```
NotSpecial: `print`?

amivapi/users/security.py
cburchert: done here and also in __init__.py

cburchert:
Fixed the problems. I recommend squashing.

PR #224, opened 2018-01-27
Move secret key from config to database.

NotSpecial:


amivapi/auth/utils.py
```diff
@@ -8,6 +8,29 @@
...
+    with app.app_context():  # App context needed for db connection
+        key = 'TOKEN_SECRET'
+        db = app.data.driver.db['config']
+        result = db.find_one({key: {'$exists': True, '$nin': [None, '']}})
+
+        if result is not None:
+            secret = result[key]
+        else:
+            # Create secret and save in db
+            secret = b64encode(urandom(32)).decode('utf_8')
```
cburchert: Not every sequence is valid utf-8, so this might fail somehow.

cburchert: ah I missed the b64encode

NotSpecial: It's first base64 encoded, which can always be utf-8 decoded.

amivapi/tests/auth/test_secret.py
```diff
@@ -0,0 +1,61 @@
...
+        super(SecretTest, self).setUp()
+
+        with self.app.app_context():
+            config_value = self.app.config.get(SECRET_KEY)
+            db_item = self.db['config'].find_one({
+                SECRET_KEY: {'$exists': True}
+            })
+
+        self.assertIsNotNone(config_value)
+        self.assertIsNotNone(db_item)
```
cburchert: Maybe test for nonempty string as well? Or just assertTrue? This could also pass an empty string otherwise.

cburchert: Also why db_item here and db_item[SECRET_KEY] in the next line?

amivapi/tests/auth/test_secret.py
```diff
@@ -0,0 +1,61 @@
...
+    def setUp(self):
+        """Skip the normal api setup."""
+
+    def test_no_secret(self):
+        """Without init_key, no secret key is in config or db."""
+        # Replace init_secret with a dummy function, which does nothing
+        with patch('amivapi.auth.init_secret'):
+            # Now call setup -- no secret will be initialized
+            super(SecretTest, self).setUp()
+
```
cburchert: The with block could end here

amivapi/auth/utils.py
```diff
@@ -8,6 +8,29 @@
 from os import urandom
 
 
+def init_secret(app):
+    """Get secret key from database or create it. Put in config.
+
+    The secret key is stored in the database to ensure consistency.
+    If no secret is in the database yet, create it.
+
+    The database collection holding this key is called `config`
```
cburchert: . is missing at the end of the sentence

amivapi/tests/auth/test_secret.py
NotSpecial: it does now :)

PR #230, opened 2018-01-27
WIP: eve 0.8 support

cburchert:
Some problems solved already, however there is an issue with the schemas that use cascade_delete.


cburchert:
Closing this for now, as we need more work before a pull request makes sense.

PR #243, opened 2018-02-25
Create mailing list files remotely via ssh.

### LINKED ISSUE #239: Make AMIVAPI Stateless Again (MASA)
I'm a little bit annoyed that we still require local files because of the mailing lists,
which complicates deployment e.g. with docker swarm or other container solutions,
since we somehow need to to share this data between replicas of the API.

Talking to @temparus I realized that this is furthermore redundant since those files need to
be copied to the mail server over ssh periodically anyways.

Which leads me to the following idea:

*Why not use `ssh` to directly create/remove the files on the mailserver?*

In the config, we can accept username and password (instead of a local path as currently done)
and using a subprocess we can easily call ssh to write the file, e.g. like
[this](https://stackoverflow.com/a/19202567)

Pros:
* Make API instances stateless, simplify deployment
* Immediate Update on Mail Server, not just after another script copies the lists

Cons:
* Slightly longer response time since the file needs to be written over ssh first?

Opinions?

Edit: We could also just use ssh keys, of course

NotSpecial:
Until now, mailing list files are created locally.
For the mailing lists to actually work, they need to be copied to a remote
server.

This copying is now included in AMIVAPI, which has two main benefits:

- No local files are required anymore, useful for future deployment with docker
- No external copying tool is required to make mailing lists work with amivapi

Unfortunately, the mailing list server supports *only* ssh connections,
therefore we have no other option to transfer the files.

While I do not think this is a very *good* solution, I believe its the *best*
approach we have available.

Closes #239.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,11 @@ Now it's time to configure AMIVAPI. Create a file `config.py`
 
     # Mailing lists for groups (optional, uncomment if needed)
     # MAILING_LIST_DIR = '/directory/to/store/mailing/list/files/'
+    
+    # Remote mailings list files via ssh (optional)
+    # REMOTE_MAILING_LIST_ADDRESS = 'user@remote.host'
+    # REMOTE_MAILING_LIST_KEYFILE = ''
+    # REMOTE_MAILING_LIST_DIR = './'
 
     # SMTP configuration for mails sent by AMIVAPI (optional)
     # API_MAIL = 'api@amiv.ethz.ch'
--- a/amivapi/cli.py
+++ b/amivapi/cli.py
@@ -12,7 +12,7 @@ from click import argument, echo, group, option, Path
 from amivapi.bootstrap import create_app
 from amivapi.cron import run_scheduled_tasks
 from amivapi import ldap
-from amivapi.groups.mailing_lists import new_groups
+from amivapi.groups.mailing_lists import update_group
 
 
 @group()
@@ -30,9 +30,11 @@ config_option = option("--config",
 def recreate_mailing_lists(config):
     """(Re-)create mailing lists for all groups.
 
-    1. Delete all mailing list files in mailing list directory.
+    1. Delete all mailing list files.
 
-    2. Create new mailing list files for each group in the database.
+    2. Create new mailing list files.
+
+    For every group, we call the update_group function for this
     """
     app = create_app(config_file=config)
     directory = app.config.get('MAILING_LIST_DIR')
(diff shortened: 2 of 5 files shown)


NotSpecial:
More code appears to have changed than actually did. I just moved some functions around which confused git a lot.

temparus:
This approach could create some problems if the files are updated at the time when a group gets updated.

Mails sent in the time between the removal and the transfer of the new forward files will be lost as there is no information for the mail server where to forward the email at this time. The sender does not get an error message in that case, so the email won't be delivered a second time.

hermannsblum:
@temparus Doesn‘t the same problem happen right now when overwriting the files?

temparus:
To minimize the risk of loosing messages, the forward files are updated every day at 3 AM only right now.

hermannsblum:
@temparus @NotSpecial As I understand, one of the objectives of the ssh stuff is to make deployment easier. Then again, I see the point that overwriting these mailfiles during daytime is dangerous. Would it be a valid option to store the files in the database instead of the local filesystem and then still only update them once per day?
This could either work with simply storing the file contents or even with GridFS

NotSpecial:
Thanks for the feedback, this is an important issue.

I am not a big fan of waiting until 3am -- I think it would be great if changes to the groups would be active as soon as possible.

I have made to small changes:

1. When *updating*, not all files are removed and re-created anymore. Only files which are not needed anymore will be removed, the other files will updated.

2. The update over ssh itself is now split in two parts: first, the file is uploaded to a temporary file. As soon as the upload is complete, the mailing list file is replaced using `mv`. I have done a little research and believe this is about [as atomic as we can replace a file on linux](https://unix.stackexchange.com/a/322074).

With this, the issue should be eliminated: At no point, a mailing list file which is still required will disappear, so no message will get lost.

hermannsblum:
@NotSpecial very good idea

hermannsblum:
@NotSpecial One question remaining for me is whether we want this possiblity of simultaneous remote and local files (possible by setting all the strings). This could be dangerous in cases where somebody sets up a local email
handling while the remote one is still in place and vice versa. However, I am not familiar enough with the overall system to say whether this really would be a problem. Still, why not simply add a config MAILLOCATION which is either ‚REMOTE‘ or ‚LOCAL‘, instead of the REMOTE prefix?

NotSpecial:
I see your point, but I am not sure if it would simplify the code or config.

Currently, the user can decide which feature is needed. Since both options are deactivated by default,
it cannot happen by accident that both local and remote files are used, it has to be intentionally configured. So I guess it is not that dangerous and would keep it as is.

PR #250, opened 2018-03-08
WIP: Change membership detection used with LDAP

temparus:
Unfortunately, no entries get deleted at the LDAP field `ou` (organizational unit). This leads to wrong detection of membership for students which have changed their field of study to one not assigned to us.

The field `departmentnumber` seems to contain just the currently assigned field(s) of study, which should fix the issue.

### DIFF
--- a/amivapi/ldap.py
+++ b/amivapi/ldap.py
@@ -74,10 +74,10 @@ def sync_all():
     Returns:
         list: Data of all updated users.
     """
-    # Create query: VSETH member and part of any of member ou
-    ou_items = ''.join(u"(ou=%s)" % _escape(item) for item in
-                       current_app.config['LDAP_MEMBER_OU_LIST'])
-    query = u"(& (ou=VSETH Mitglied) (| %s) )" % ou_items
+    # Create query: VSETH member and departmentnumber of any member
+    departmentnumber_items = ''.join(u"(departmentnumber=%s)" % _escape(item) for item in
+                       current_app.config['LDAP_MEMBER_DEPARTMENTNUMBER_LIST'])
+    query = u"(& (ou=VSETH Mitglied) (| %s) )" % departmentnumber_items
     ldap_data = _search(query)
 
     return [_create_or_update_user(user) for user in ldap_data]
@@ -91,6 +91,7 @@ def _search(query):
         'givenName',
         'sn',
         'swissEduPersonGender',
+        'departmentnumber',
         'ou'
     ]
     results = current_app.config['ldap_connector'].search(query,
@@ -129,17 +130,20 @@ def _filter_data(data):
     res['gender'] = \
         u"male" if int(data['swissEduPersonGender']) == 1 else u"female"
 
-    if 'D-ITET' in data['ou']:
+    if any('D-ITET' in item for item in data['departmentnumber']):
         res['department'] = u"itet"
-    elif 'D-MAVT' in data['ou']:
+    elif any('D-MAVT' in item for item in data['departmentnumber']):
         res['department'] = u"mavt"
     else:
         res['department'] = u"other"
 
-    # ou contains all 'organization units'. This contains fields of study.
-    # Check if it contains any field of study assigned to us
-    ou_list = current_app.config['LDAP_MEMBER_OU_LIST']
-    is_member = bool(set(data['ou']).intersection(set(ou_list)))
+    # departmentnumber contains the list of the current field of study.
+    # Check if is contains any value assined to us
+    department_list = current_app.config['LDAP_MEMBER_DEPARTMENTNUMBER_LIST']
+    is_member = False
+    for search_str in department_list:
+        if any(search_str in item for item in data['departmentnumber']):
+            is_member = True
 
     if ('VSETH Mitglied' in data['ou']) and is_member:
         res['membership'] = u"regular"
--- a/amivapi/settings.py
+++ b/amivapi/settings.py
@@ -97,6 +97,14 @@ CONFIRM_EMAIL_TEXT = (
 CONFIRM_TEXT = "Your singup was confirmed!"
 SIGNUP_DELETED_TEXT = "Your signup was removed."
 
+# All department affiliations in ldap (departmentnumber) which are assigned to AMIV (by VSETH)
+LDAP_MEMBER_DEPARTMENTNUMBER_LIST = [
+    u'ETH Student D-ITET',
+    u'ETH Studentin D-ITET',
+    u'ETH Student D-MAVT',
+    u'ETH Studentin D-MAVT'
+]
+
 # All organisational units (ou) in ldap which are assigned to AMIV (by VSETH)
 LDAP_MEMBER_OU_LIST = [
     u'Biomedical Engineering MSc',


NotSpecial:
I have updated the tests for LDAP, cleaned up to required entries in settings and used the skipping feature of pytest (used by the ssh integration tests so far) for the ldap integration tests as well to integrate them better into the test suite.
I also improved the README to explain how to run LDAP (and ssh) integration tests

README.md
```diff
@@ -178,6 +182,33 @@ To run just one python version:
 tox -e py36
 ```
 
+### Integration Tests
+
+The integration tests for ssh mailing list creation and LDAP require additional
+information to be run, which should not be published.
+By default, these tests are *skipped*, as you can see in the tox summary.
+
+#### SSH
```
hermannsblum: What does SSH have to do with this? What kind of SSH connection is required here?

amivapi/cli.py
```diff
@@ -90,7 +90,7 @@ def ldap_sync(config, sync_all, nethz):
             else:
                 for user in nethz:
                     if ldap.sync_one(user) is not None:
-                        echo("Succesfully synchronized '%s'." % user)
+                        echo("Successfully synchronized '%s'." % user)
```
hermannsblum: I think this is the most important change of this PR, well done 👍

amivapi/ldap.py
```diff
@@ -28,9 +28,10 @@
 
 def init_app(app):
     """Attach an ldap connection to the app."""
-    user = app.config['LDAP_USER']
-    password = app.config['LDAP_PASS']
-    app.config['ldap_connector'] = AuthenticatedLdap(user, password)
+    user = app.config['LDAP_USERNAME']
+    password = app.config['LDAP_PASSWORD']
+    if user and password:
```
hermannsblum: What does this `if` actually do?
The only case I see where this becomes relevant is in case `LDAP_USERNAME` and `LDAP_PASSWORD` are actually set, but evaluate to some statement that is false (empty strings, `False`, ...). This would be a weird case, but it would be skipped quite intransparantly to the user (because nothing would happen, but LDAP would also not work).
Is there any other case I do not see which is a reason for this `if` statement?

amivapi/ldap.py
```diff
@@ -129,22 +133,15 @@ def _filter_data(data):
...
-    # ou contains all 'organization units'. This contains fields of study.
-    # Check if it contains any field of study assigned to us
-    ou_list = current_app.config['LDAP_MEMBER_OU_LIST']
-    is_member = bool(set(data['ou']).intersection(set(ou_list)))
-
-    if ('VSETH Mitglied' in data['ou']) and is_member:
-        res['membership'] = u"regular"
-    else:
-        res['membership'] = u"none"
+    # The departmentNumber field contains certain phrases for each department
```
hermannsblum: While I find the generator-style search very elegant, I think a bit further explanation on what the departmentNumber is and how this search works would be helpful to prevent wrong maintainance changes in the future to this very crucial part of the LDAP synchronisation.

amivapi/ldap.py
```diff
@@ -129,22 +133,15 @@ def _filter_data(data):
...
+    department_map = current_app.config['LDAP_DEPARTMENT_MAP'].items()
+    department = (dept for phrase, dept in department_map
+                  if phrase in data['departmentNumber'][0])
+    res['department'] = next(department, None)  # None if no match
+
+    # Membership: One of our departments and VSETH member
+    is_member = res['department'] and ('VSETH Mitglied' in data['ou'])
```
hermannsblum: I think pytest would complain here to make it more explicit and write `res['department'] is not None and ...`

amivapi/settings.py
```diff
@@ -97,30 +96,19 @@
...
-    u'Robotics, Systems and Control MSc',
-    u'Verfahrenstechnik MSc',
-    u'Doktorat Informationstechnologie & Elektrotechnik',
-    u'Doktorat Maschinenbau und Verfahrenstechnik'
-]
+
+# In LDAP, the 'departmentNumber' field contains the students departments
+# It contains: type of person, e.g. student, department, exact field of study
+# We can use this to discover department of students
+LDAP_DEPARTMENT_MAP = {
```
hermannsblum: Is this also tested with mobility students?

temparus: Yes, it also works with mobility students.

README.md
NotSpecial: It's just a bit tangential. Since I updated the integration tests for ldap and needed to add that to the Readme, I included the ssh integration tests here as well while I was at it. No functionality changes.

amivapi/cli.py
NotSpecial: #proud

amivapi/ldap.py
NotSpecial: The behavior is intended, if the strings are empty (no username and/or password), the eth ldap won't work.

The fields are never not set, since `settings.py` contains defaults.

I could change the defaults to `None`, but not sure if that would help.

I think I will add some logging to inform the user whether ldap was enabled.

amivapi/ldap.py
NotSpecial: Good point, I think I will simply add a concrete example for the departmentNumber field to avoid confusion.

amivapi/ldap.py
hermannsblum: What would be the problem if ldap simply fails because of empty strings? I feel like that would be more transparent behaviour

NotSpecial: Ok, I have changed the defaults to None and use whatever value specified if both username and password are not None

amivapi/ldap.py
```diff
@@ -12,10 +12,26 @@
...
-- The `_filter_data' function is rather complicated. Maybe some parts could
+- The `_process_data' function is rather complicated. Maybe some parts could
   be improved or moved to the nethz module?
 - `_create_or_patch_user' is also not very straightforward, maybe the ldap
   importing logic could be simplified?
+
+
+Note on department info in ldap:
+
+We can discover a users department by looking at the `departmentNumber` field.
```
cburchert: user's

amivapi/ldap.py
```diff
@@ -12,10 +12,26 @@
...
+Contrary to it's name, it does not contain a number, but entries like this:
```
cburchert: its

amivapi/ldap.py
```diff
@@ -12,10 +12,26 @@
...
+
+- `ETH Studentin D-ITET, Elektrotechnik und Interformationstechnologie Bsc.`
+- `ETH Student D-MAVT, Doktorand`
+
+and similar. For regular, mobility and phd students the starting part is
+always the same, and the second part of the string contains further details.
+
+We therefore match the first part of the `departmentNumber` field to
+check for the department. In the app settings, we define which substring
+is mapped to which deparment.
```
cburchert: department

cburchert:
LGTM.

I think we could do something to make the remote SSH tests easier to run. There are very simple SSH servers or SSH server mockers out there. I think this would be something we can just include in the repo. Another way would be to configure travis to run the SSH tests. That would prevent undetected accidental breaking of the SSH stuff.

Both is certainly nothing that must go into this pull request. The existing tests are very good anyway, so maybe we have higher priority stuff.

PR #257, opened 2018-03-18
Set password in PATCH response to None (Fixes #216)

### LINKED ISSUE #216: Cached password sent with PATCH response
The cached password is sent with the user object returned after a successful `PATCH` to `/users/<user-id>`.

The property `password` should be removed from the response.

temparus:
This is not the perfect solution as the `password` field is still sent with the PATCH response. It gets cleared at least before sending the data to the client.


temparus:
Switched to `on_post_<method>_users` hook for all used methods. This way, we can make sure that the fields are actually removed from the response.

amivapi/users/__init__.py
```diff
@@ -32,3 +32,8 @@ def init_app(app):
     app.on_updated_users += project_password_status_on_updated
     app.on_replace_user += hash_on_update
     app.on_replaced_user += project_password_status_on_updated
+
+    # on_post_METHOD, triggered before sending the response by Eve
+    for method in ['GET', 'POST', 'PATCH', 'PUT']:
+        event = getattr(app, 'on_post_' + method + '_users')
+        event += hide_fields
```
NotSpecial: Regarding the current travis error:

Your new hooks should work, but the new implementation of `hide_fields` crashes the old hooks, which are still present.

You need to remove these lines above:
```
app.on_fetched_item_users += hide_fields
app.on_fetched_resource_users += hide_fields
```
(Deprecated anyway since covered by the `on_post` hook)

Then everything should work.

temparus: Unfortunately, there are still errors with the Travis build.

NotSpecial: I will investigate.

In the mean time I have noticed something else: You can remove the `PUT` from methods. AMIV API does not support `PUT~

temparus: Are you sure?

The following line indicates something else:
https://github.com/amiv-eth/amivapi/blob/149d3ebaaea1680a8a566f80fe7d604843343374/amivapi/settings.py#L35

NotSpecial:
This was a more general problem we had already in another place. The `on_post` hook only receives a `Response` object -- one has to extract the data out of it.

I have now written a generic wrapper to make the existing solution re-usable.

Unfortunately, there is still an issue:

On embedded resources, passwords are not hidden anymore. Our current method handling embedded resources (in `utils.py`) only recursively applies `on_fetched` hooks. I am not sure if this can be extended to the `on_post` hooks

NotSpecial:
I started to extend the current implementation, but it fails for the following reason:
The `on_post` hooks require an `Reponse` objet, which takes a string as argument.
Unfortunately, the data we have is not easily serializable using json since things like `ObjectId` can not be handled.

Eve is doing this internally quite complicated. in `methods.common` there actually is a `build_response_document` function, but when I tried to use it I only got `None` in return.

I'm a bit at a loss here. Any ideas what we could do @cburchert ?

temparus:
Did you find a solution for this issue?

amivapi/users/security.py
```diff
@@ -78,31 +79,55 @@ def create_user_lookup_filter(self, user_id):
...
+        request, response: unused
+        payload (dict): response data
+    """
+    # Use either the '_items' field (resource requests) or the
+    # whole payload as one item (item requests)
+    for item in payload.get('_items', [payload]):
+        _hide_fields(item)
+
+
+def hide_after_fetch(item):
```
cburchert: Do we need this function? We could just use _hide_fields directly instead.
Or if we want the extra name we could do:

```
hider_after_fetch = _hide_fields
```

PR #259, opened 2018-03-18
Fix event queue

cburchert:


amivapi/tests/events/test_queue.py
```diff
@@ -59,6 +57,23 @@ def test_fcfs_users_get_auto_accepted(self):
...
+        user = self.new_object('users')
+
+        self.api.post('/eventsignups', data={
+            'user': str(user['_id']),
+            'event': str(event['_id'])
+        }, status_code=201)
+
+        accepted = self.api.get('/eventsignups?where={"accepted":true}',
+                                status_code=200).json['_items']
+        self.assertEqual(len(accepted), 1)
```
NotSpecial: I think it would be more straightforward to check for the particular signup we created and check if *this* signup is accepted, instead of checking if there is one accepted signup in total.

NotSpecial: I will quickly add this.

NotSpecial:
I adjusted the tests accordingly: Now we immediately check the signup response whether it got accepted or not.

Because of this I discovered that newly created singups would always contain `accepted: False`, even if they got, because the response was never updated. So I implemented this.

While doing that, I remove some hook. Eve does not have `on_insert_item` hooks ([reference](http://python-eve.org/features.html#database-event-hooks)), I think that was changed some versions ago.

But because of the weird way Eve registers hooks, it does not complain about non-existing hook (great design, really...) so this never caused an issue. I cleaned them up :)

@cburchert, please have a look at my changes, afterwards you can merge.

Issue #235, opened 2018-02-10
Changing the Password


Moschn:
Changing the password should also require the current password. Currently an attacker could steal a session token and then change the password and thus hijack the account.

cburchert:
Maybe an easy way to implement this would be to require a session of a very
small maximum age to change the password (say 1 minute).

2018-02-10 17:02 GMT+01:00 Moritz Schneider <notifications@github.com>:

temparus:
I like the approach of @cburchert. Other opinions? @NotSpecial @hermannsblum

temparus:
This issue has been resolved with PR #260.

PR #260, opened 2018-03-18
Add session_younger_than validator

temparus:
Add validation rule to force the token to be younger than a given threshold.

Fixes issue #235.


NotSpecial:
I think it is unintuitive to implement this functionality as a validator, since it has nothing to do with
the validity of the password field.

I would rather include this in e.g. the user auth class or as a `on_pre-_PATCH` hook and abort with an appropriate http status code instead of an validation error.

amivapi/validation.py
```diff
@@ -121,3 +122,22 @@ def _validate_filetype(self, filetype, field, value):
...
+        Validates if the session is not older than threshold_time
+
+        Except admins, they can ignore this
+
+        Args:
+            threshold_time (int): threshold to to compare with in seconds
+            field (string): field name.
+            value: field value.
+        """
+        time_created = g.current_session['_created'].replace(tzinfo=None)
```
NotSpecial: you should put an `if not g.get('resource_admin'):` around everything so this really only applies to non-admins.

amivapi/validation.py
```diff
@@ -121,3 +122,22 @@ def _validate_filetype(self, filetype, field, value):
...
+        time_now = datetime.now()
+
+        if threshold_time > 0 and time_now - time_created > threshold_time:
```
cburchert: I am a bit worried about the timezones here. datetime.now() returns a local time, but as far as I remember we want UTC everywhere (and even enforce UTC in the external API).
How about:

```
time_created = g.current_session['_created']
time_now = datetime.now(timezone.utc)
```

Then the comparison should be safe.

amivapi/validation.py
```diff
@@ -121,3 +122,22 @@ def _validate_filetype(self, filetype, field, value):
         if not(t in filetype):
             self._error(field, "filetype '%s' not supported, has to be in: "
                         "%s" % (t, filetype))
+
+    def _validate_session_younger_than(self, threshold_time, field, value):
```
cburchert: As the units of ```threshold_time``` are very important here, I would recommend including it in the parameter name or even better in the function name. How about ```_validate_session_younger_than_s```?

amivapi/validation.py
temparus: This is a great idea as far as the times saved in the database are also in UTC.

NotSpecial: The easiest way would be using`datetime.utcnow()`, I guess

amivapi/validation.py
NotSpecial: I would leave the function name as is (in the end, you can pass whatever you want anyway), but defintitely include it in the validators docstring.

In particular, I would not use an integer argument at all, since it really is ambiguous when it comes to time units. I would use a `datetime.timedelta`, since it is
- more descriptive
- more flexible

e.g. you could use it like:
```
from datetime import timedelta
(...)
'session_younger_then': timedelta(seconds=1)
```

What do you think? Afaik it should work without any changes, since we are using differences of time objects anyway -- they can be compared to timedeltas by design

amivapi/users/model.py
```diff
@@ -99,7 +99,8 @@
                 'nullable': True,
                 'default': None,
                 'description': 'Leave empty to use just LDAP authentification. '
-                'People without LDAP should use this field.'
+                'People without LDAP should use this field.',
+                'session_younger_than': 60
```
NotSpecial: See disucssion: use `timedelta(seconds=60)` or `timedelta(minutes=1)` here

amivapi/validation.py
```diff
@@ -121,3 +122,22 @@ def _validate_filetype(self, filetype, field, value):
...
+        """Validation of the used token for special fields
+
+        Validates if the session is not older than threshold_time
+
+        Except admins, they can ignore this
+
+        Args:
+            threshold_time (int): threshold to to compare with in seconds
```
NotSpecial: See discussion: Don't specify `int`, but `timedelta` as argument type

amivapi/validation.py
cburchert: datetime.utcnow does not have a timezone attached. That's why I prefer the other way ;)

amivapi/validation.py
cburchert: yes good idea. Definitively go for timedelta.

cburchert:
I implemented the timedelta proposal and completed the test. Because I used the python 3 timezone class instead of pytz, this code does not work in python 2 anymore. The tests should work after we merge the oauth branch.

I see one more problem: Using this validator on the password field prevents admins from creating users with passwords.

cburchert:
I realized this is already handled in the validator. Added a test.

temparus:
Python 2 support should be dropped, so we can merge this PR.

PR #270, opened 2018-05-08
Stop non-admins from querying restricted fields.

NotSpecial:
Fields like legi number are hidden from non-admin users.
Nevertheless, it was possible to query for those fields, e.g. for a
particular legi number, which will only return a single result.

And even if the legi number is hidden, since it is unique it can be
matched to its user without problem.

Changes:
- Eve already allows restricting query parmeters, which is now done
  dynamically for user requests
- Test included


NotSpecial:
Yeah, I see your point -- I am not too happy as well, because changing the config for the processing of a single request is not really a clean solution.

PR #287, opened 2018-08-21
Set minlength for password

### LINKED ISSUE #234: Password Policy
There should be a password policy. We already use a good hash function for passwords but a password with length 1 is currently allowed and easily broken. 

I propose just a minimum length of 7 for passwords containing only letters. This would lead to a searching space of 24^7 and with a required time of around 100ms per hash it would take 14.5 years.

In order to prevent password list attacks, a more complicated policy with alphanumeric and sepcial character requirements could also be useful.

temparus:
Implements the minimal password policy discussed in issue #234 .

Resolves #234.

### DIFF
--- a/amivapi/users/model.py
+++ b/amivapi/users/model.py
@@ -96,6 +96,7 @@ userdomain = {
             # Fields the user can modify himself
             'password': {
                 'type': 'string',
+                'minlength': 7,
                 'maxlength': 100,
                 'empty': False,
                 'nullable': True,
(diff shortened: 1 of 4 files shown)


amivapi/users/model.py
```diff
@@ -96,6 +96,7 @@
             # Fields the user can modify himself
             'password': {
                 'type': 'string',
+                'minlength': 7,
```
NotSpecial: Any particular reason why it's 7?

temparus: It was the proposal from the linked issue. It does not necessarily need to be 7, but I think this reasonable.

NotSpecial: Fair enough!

PR #285, opened 2018-07-14
Update requirements.txt

gartens:
As there are currently issues with Flask-PyMongo 2.0.0, add a explicit dependency for Flask-PyMongo 0.5.2.
See pyeve/eve#1172.

PR #288, opened 2018-08-21
Pin Flask-PyMongo version to prevent incompatible update to version 2.0

cburchert:


NotSpecial:
There is already another pull request regarding this issue: https://github.com/amiv-eth/amivapi/pull/285

PR #289, opened 2018-08-21
Docker cron

### LINKED ISSUE #275: Extend docker image with Cron Mode
We should extend the the docker image with a Cron Mode, which can be enabled by an environment variable.

This allows us to spin up multiple instances of the API for serving requests and one single container which executes all the regular tasks.

temparus:
Adds another Dockerfile to create a docker image which is dedicated to running cron.

Resolves #275.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,12 @@ docker service create \
     --name amivapi  -p 80:8080 --network backend \
     --secret amivapi_config \
     amiveth/amivapi
+
+# Create new API cron service with secret
+docker service create \
+    --name amivapi-cron  --network backend \
+    --secret amivapi_config \
+    amiveth/amivapi-cron
 ```
 
 If you want to use a different name for the secret (or cannot use secrets
--- a/cron.Dockerfile
+++ b/cron.Dockerfile
@@ -1,0 +1,27 @@
+FROM python:3.6-alpine
+
+# Create user with home directory and no password and change workdir
+WORKDIR /api
+# Environment variable for config, use path for docker secrets as default
+ENV AMIVAPI_CONFIG=/run/secrets/amivapi_config
+
+# Install bjoern and dependencies for install (we need to keep libev)
+RUN apk add --no-cache --virtual .deps \
+        musl-dev python-dev gcc git
+
+# Copy files to /api directory, install requirements
+COPY ./ /api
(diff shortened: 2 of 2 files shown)


cron.Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+# Install bjoern and dependencies for install (we need to keep libev)
+RUN apk add --no-cache --virtual .deps \
+        musl-dev python-dev gcc git
+
+# Copy files to /api directory, install requirements
+COPY ./ /api
+RUN pip install -r /api/requirements.txt
+
+# Cleanup dependencies
+RUN apk del .deps
```
NotSpecial: Everything above seems exactly like the normal dockerfile.

Could you not simply use FROM amiveth/amivapi instead to use the normal Dockerfile as starting point to reduce maintenance, if we even need to change something?

README.md
```diff
@@ -52,6 +52,12 @@ docker service create \
     --name amivapi  -p 80:8080 --network backend \
     --secret amivapi_config \
     amiveth/amivapi
+
+# Create new API cron service with secret
```
NotSpecial: Some little explanation what the cron service is for might be useful.

NotSpecial:
As both Dockerfiles are nearly identical, I have merged them into one.

Two separate startup scripts run the container either in server mode (default) or cron mode.
The README is adjusted accordingly.

However, I discovered something during those changes:
*cron does not work!*

First of all, the wrong script was executed. I fixed that, and now correctly `amivapi cron` is executed. However, this doesn't seem to solve the issue and I'm at a loss now.

Things I have considered for the crontab:

- Use only absolute pathnames for directory (`crond` can't find the custom directory otherwise) and script paths
- Username as crontab filename
- Maximum verbosity for logging

Yet the script does not seem to be executed and there is no output at all. No idea how to debug this further.

NotSpecial:
Cron is gone!

Since we have scheduled tasks for a specific time anyway, we do not need cron to run only at specific times.

As a result, I have simply modified the existing CLI to be able to run in `cron` - mode, executing scheduled tasks periodically.

NotSpecial:
@cburchert , @temparus , could you have a quick look at my changes?

I have (in my opinion) improved the README a little, since I had to update it anyways for the docker commands.

PR #290, opened 2018-08-22
Add write restrictions for eventsignups

### LINKED ISSUE #240: Restrict event self signoff
We should restrict self signoff from participants of an event before the event starts.

Possible solutions are:

* Only allow self signoff within the registration window.
* Add an additional optional field to the event entity `signoff_end` to specify until when the users are allowed to signoff themselves.

temparus:
Allows write access to `/eventsignups` for non-admins only within the registration window.
Admins can write to the resource at all time.

Write access means `POST`, `PATCH` and `DELETE`.

Resolves #240.

### DIFF
--- a/amivapi/events/authorization.py
+++ b/amivapi/events/authorization.py
@@ -5,7 +5,7 @@
 """Authorization for events and eventsignups resources"""
 
 from flask import g, current_app
-from datetime import timezone, datetime as dt
+from datetime import timezone as tz, datetime as dt
 from amivapi.auth import AmivTokenAuth
 
 
@@ -24,8 +24,10 @@ class EventSignupAuth(AmivTokenAuth):
             lookup = {current_app.config['ID_FIELD']: item['event']}
             event = current_app.data.find_one('events', None, **lookup)
 
-        return event['time_register_start'].replace(tzinfo=timezone.utc) <= dt.utcnow()
-            <= event['time_register_end'].replace(tzinfo=timezone.utc)
+        time_register_start = event['time_register_start'].replace(tzinfo=tz.utc)
+        time_register_end = event['time_register_end'].replace(tzinfo=tz.utc)
+
+        return time_register_start <= dt.utcnow() <= time_register_end
 
     def has_resource_write_permission(self, user_id):
         """Anyone can sign up. Further requirements are enforced with validators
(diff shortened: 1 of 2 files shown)


amivapi/events/authorization.py
```diff
@@ -15,9 +15,20 @@ def create_user_lookup_filter(self, user_id):
...
+        if isinstance(item['event'], dict):
+            event = item['event']
+        else:
+            # Event is not embedded, get the event first
+            lookup = {current_app.config['ID_FIELD']: item['event']}
+            event = current_app.data.find_one('events', None, **lookup)
+        time_now = datetime.now(timezone.utc)
+
+        if time_now - event['time_register_start'] > timedelta(seconds=0) and \
+           time_now - event['time_register_end'] < timedelta(seconds=0):
```
NotSpecial: Why the substraction business instead of `if event['time_register_start] <= time_now <= event['time_register_end']`? (yep, the double `<` works in python btw^^)

NotSpecial: So actually this could be shortened to:

```
return event['time_register_start] <= time_now <= event['time_register_end']
```

NotSpecial: Furthermore, `datetime` includes the `utcnow` method, shortening the above to: 

```
return event['time_register_start] <= dt.utcnow() <= event['time_register_end']
```

amivapi/events/authorization.py
```diff
@@ -4,8 +4,8 @@
 #          you to buy us beer if we meet and you like the software.
 """Authorization for events and eventsignups resources"""
 
-from flask import g
-
+from flask import g, current_app
+from datetime import datetime, timedelta, timezone
```
NotSpecial: ```from datetime import datetime as dt```

(see below)

NotSpecial:
Thanks for the contribution, in particular for adding tests, too :)

I have some (small recommendations) regarding the style of the checks.

NotSpecial:
Huh, there is a weird error now regarding "timezone aware and non timezone aware datetimes", (see travis).
Do you have any idea what the issue could be?

temparus:
Hm, maybe the following could help?
https://stackoverflow.com/a/7065242/4581090

temparus:
The issue with the timezones seems to be resolved now.

PR #294, opened 2018-09-28
Eve08

NotSpecial:
Eve 0.8 brings Cerberus 1.0, which requires lots of changes:

- Several properties have been renamed
- type checking is now handled differently
- validation rules require a schema in the docstring to validate the schema
  itself

This patch updates AMIV API to work with Cerberus 1.0.

### DIFF
--- a/amivapi/events/validation.py
+++ b/amivapi/events/validation.py
@@ -8,7 +8,7 @@ from datetime import datetime
 import json
 
 from flask import current_app, g, request
-from jsonschema import Draft4Validator, SchemaError, ValidationError
+from jsonschema import Draft4Validator, SchemaError
 import pytz
 
 
@@ -33,7 +33,7 @@ class EventValidator(object):
 
         try:
             data = json.loads(value) if value else {}  # Do not crash if ''
-        except json.JSONDecodeError:
+        except json.JSONDecodeError as e:
             self._error(field,
                         "Must be json, parsing failed with exception: %s" % e)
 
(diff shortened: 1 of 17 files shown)


amivapi/events/validation.py
```diff
@@ -138,68 +145,81 @@ def _validate_email_signup_must_be_allowed(self, enabled, field, value):
...
-        """Retrieve time field from document or _original_document."""
+        """Retrieve time field from document or original document."""
         # Try to pick the value from document first, fall back to original
         time = self.document.get(fieldname)
         if time is None:
-            time = (self._original_document[fieldname]
-                    if self._original_document else None)
-
-        if time is None:
-            return None
```
cburchert: Why did you remove these two line? If time is now None here, this will crash the API as None does not have a replace function, which is still called in the next line.

amivapi/events/validation.py
```diff
@@ -138,68 +145,81 @@ def _validate_email_signup_must_be_allowed(self, enabled, field, value):
...
+    def _validate_json_schema(self, enabled, field, value):
+        """Validate a json schema[1] string.
 
-        1.  Is it JSON?
+        1.  Is the string valid JSON?
         2.  Does it satisfy our restrictions for jsonschemas?
         3.  Is it a valid json-schema?
 
         Args:
-            field (string): field name.
```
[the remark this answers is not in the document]
NotSpecial: Oh yes, this is a mistake: At first, I tried to use the new cerberus types, which to not have `field` anymore, so I removed it. Later I decided that using a `string` type and additional validator works better with our online docs, so I changed the function, but forgot to update the docstring.

Will correct :)

amivapi/events/validation.py
NotSpecial: I merged it into the return statement: `return time.replace(tzinfo=None) if time is not None else None`

Do you think this looks too confusing? Then I can change it back^^

PR #301, opened 2018-10-13
Remove admin privileges for session_younger_than validator

temparus:
Admins should not have special privileges with the `session_younger_than` validator. So administrators should also comply with this validator.

Otherwise, if someone steals the token from an administrator, a password for any user can be set.

### DIFF
--- a/amivapi/tests/test_validation.py
+++ b/amivapi/tests/test_validation.py
@@ -56,7 +56,7 @@ class ValidatorAMIVTest(WebTest):
         self.new_object("groupmemberships",
                         user=user['_id'], group=admin_group['_id'])
 
-        # User is now admin, so can always post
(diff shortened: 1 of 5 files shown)


NotSpecial:
The changes look good, however the tests now fail.

As admins have no exception anymore, the tests cannot create sessions anymore.

This got me thinking: Do we need this validator for `POST`? I mean, what we want to avoid is someone changing a users password. This does not really apply to creating new users, right?

We could use this validator only if a document gets updated, which would fix the test issue.

temparus:
For admins, this restriction does not make sense as they can just remove and recreate a user.

For normal users, which do not have the rights to recreate a user account, this false security aspect is not given anymore. There it still makes sense to keep the `session_younger_than` validator as it really improves the security.

PR #304, opened 2018-10-13
Image aspect ratio validation and no_html validation for event text

meta7x:


NotSpecial:
Rebased version merged #306

PR #306, opened 2018-10-13
Add validation for event description and image aspect ratio

### LINKED ISSUE #96: Define event image resolution, validate this somehow?
We have three image fields in events

img_thumbnail
img_web
img_infoscreen

We should specify the resolution and maybe enforce it in the validator

### LINKED ISSUE #127: Specify content of description field(and maybe others)
We just had a discussion about what the event description field contains. Our quick result was that markdown would be a good option, as everyone could generate their own formating based on the environment in which the event is rendered.
It should be considered, that newsletter entries do not have much space, therefore big amounts of text are not necessary, at the same time it would be desireable to be able to include links and simple text formating.
In any case it should be specified in the docs and if possible validated.
Another important thing here is XSS. We need to either filter HTML tags or make sure developers know that the api does not prevent the inclusion of HTML tags in fields.

NotSpecial:
Resolves #96 
Resolves #127


NotSpecial:
I see your point. The parser used in the validator would detect your example as a possible html tag, though, and subsequently not allow it. Which is what we want, because as you say: A markdown formatter would "swallow" it, resulting in unexpected output with parts of the text missing.

So I think this solution actually solves our problem.

cburchert:
Ok the parser is better than I thought. However this is not found and is a tag and disappears in my chrome:

```
Ich habe einen Party Hut auf <XD Aber no hate, die Party wird awesome :->
```

NotSpecial:
API > Chrome

<XD :-> <3

PR #313, opened 2018-11-12
Polish resource schemas and update documentation

### LINKED ISSUE #119: Documentation of Use Cases and Reasons Behind API and Data fields
During the discussion about #117 it occurred to us that a lot of previous discussions got forgotten about why we introduced certain api fields.  
It is therefore necessary that we document the reason and function behind every api field in 2 ways:
1. the public documentation of api fields  -> /docs
2. the user and/or admin documentation for more technical background and all data fields

However, we need to agree on a format for this

### LINKED ISSUE #137: Authentication example wrong in docs
The docs/User_guide.md lists 
`POST /sessions?user=mynethzormail&password=mypassword` as an example for authentication, which does not work as far as I know.

### LINKED ISSUE #168: Errors in online documentation
I just noticed this while browsing our new [online docs](http://amiv-api.ethz.ch/docs#tag/User%2Fpaths%2F~1users%2Fpost)

If you follow the link you can see that for POST, the "_id" field is part of the parameters provided.
This is incorrect, since for POST requests the server determines the id and not the client. More practically, the api would reject this.

Does anyone have an idea why this happens? A problem in Eve-Swagger, maybe?

NotSpecial:
This commit greatly adds functionality to the online documentation and
generally improve the documentation of all resources.

While I have updated the `required` and `default` properties of some fields,
as far as I am aware there should be no breaking changes.

I took this as a chance to clean up the schemas and fix some inconsistencies,
which I'll explain below.

General updates:

- Moved the DocValidator, which defined dummy rules for description etc.
  into the main validator. Rationale: Whether or not the online docs are
  activated, the descriptive schema is useful. Furthermore, deactivating the
  documentation should not cause problems with all schemas (because of missing
  validation rules).
- The API now adds definitions for all internal fields to the schema, i.e.
  _etag and so on. Eve already did this for the _id field, so I extended it
  (mainly such that the online documentation can include it nicely).
- To be consitent with item titles, resource titles are now capitalized too.

Schema updates:

- Enforced more consistency with `required`- and `default`-values. Every field
  now has to be either required xor needs a default value. If the field has a
  default value, there is no need for it to be required, but if it hasn't, the
  user must provide it. This is mainly done to remove confusion around
  missing values. This is supported by Eve (and MongoDB), but serves no
  practical use for use, yet can create confusion.
- Updated the `required` status of fields appropriately, added defaults where
  needed.
- Re-worked all field and resource descriptions to cover (hopefully) all
  functionality. Also added examples for all fields.
- In particular, added extensive descriptions behind the purpose of each
  resource, integrating the previous effort on the 'API purpose documentation'.
- For some resources like `oauthclients`, added proper resource and item titles
  (e.g. `OAuth Client`).

Documentation updates:

- Removed outdated developer and admin documentation.
- Moved documents such as the Cheatsheet or the OAuth user guide into the
  online docs, which are now the 'single source of truth' for users.
- Extended the Cheatsheet to cover sending data with multipart/form-data
- Added general introductions to REST and API auth.
- In the online docs, added support for:
  - Query parameters.
  - Authorization header.
  - Additional lookup path for GET.
  - Error Responses.
- Hooks are not displayed anymore, as they are only relevant for API developers
  and not for users.
- Updates ReDoc to version 2.

Regarding the documentation updates, I briefly considered putting them into
a pull request to Eve-Swagger. However, development on Eve-Swagger seems stale,
with two open pull requests without response. Furthermore, most additions
include AMIV API specific detail (e.g. a link to the cheatsheet), so I opted
for a API specific solution instead.

This commit resolves the last open issues for version 2.0.0:

Resolves #119
Resolves #137
Resolves #168


NotSpecial:
Btw: Does someone understand why travis fails? It seems to be a tox error. I had the same error locally, but after a small update to tox.ini it worked again. Do I need to modify travis.yml as well?

Sanity check: Does tox work for you with the current config?

NotSpecial:
Closed in favour of multiple smaller PRs.

PR #314, opened 2018-11-13
Make link methods compatible with custom resource titles

NotSpecial:
Previously, link methods would check the `title` field to identify a link
resource, which does not work if an alternative title is used.

This issue can be resolved by using the `href` field, as the key in the `DOMAIN` dict equals the endpoint URL.


cburchert:
Looks still weird but not worse than before ;)

NotSpecial:
#ShameOnMe

PR #322, opened 2018-11-13
Move schema description validation rules into general validator

NotSpecial:
We use additional schema rules such as 'description' to make our
schema more descriptive. However, Cerberus will complain about
unknown schema rules, so they have to be defined somewhere,
even if they don't do anything.

Currently, this is done in the documentation sub-module.
This is  a bit problematic, as our API is designed with replaceable
modules in mind -- but deactivating or replacing the doc module would
now cause errors in *all* schemas.

As the additional description rules are important whether or not we
use the online docs, I have moved the rules into the general validator.

### DIFF
--- a/amivapi/groups/model.py
+++ b/amivapi/groups/model.py
@@ -115,7 +115,7 @@ groupdomain = {
 
         # Additional lookup: Since group name is unique, you can use it as url
         'additional_lookup': {
-            'url': 'regex("[\w]+")',
+            'url': r'regex("[\w]+")',
             'field': 'name'
         },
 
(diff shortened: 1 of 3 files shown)


amivapi/groups/model.py
```diff
@@ -115,7 +115,7 @@ def create_user_lookup_filter(self, user_id):
 
         # Additional lookup: Since group name is unique, you can use it as url
         'additional_lookup': {
-            'url': 'regex("[\w]+")',
+            'url': r'regex("[\w]+")',
```
temparus: Why is this string prefixed as regex?

NotSpecial: This is actually changed in a separate pull request (#318).

This change is only a small fix for flake8, so we can merge this branch. (which the other PR requires)

NotSpecial: And I'm not sure what you mean with "prefixed as regex". In python the *r* makes a string *literal*, i.e. without control characters

temparus: Ah then I misunderstood the meaning of the _r_ prefix.

PR #324, opened 2018-11-13
Add test to enforce schema consistency

NotSpecial:
Currently, Eve allows fields to be undefined, which basically
only result in confusion.

To be more consistent in general, this test will enforce
that all fields are required XOR have a default value.


amivapi/tests/test_schemas.py
```diff
@@ -0,0 +1,42 @@
...
+        mark it as required and vice versa.
+        """
+        no_required_or_default = []
+        required_and_default = []
+        for resource, domain in self.app.config['DOMAIN'].items():
+            for field, definition in domain['schema'].items():
+                if definition.get('readonly'):
+                    continue
+
+                if (('required' not in definition) and
```
cburchert: You should use definition.get('required', False) 
etc instead. Currently I can make a resource:

'resource': {
  'field': {
    required: False,
    type: ...
  }
}

NotSpecial: Good point, thanks.

NotSpecial: Fixed and added comment to explain the issue.

PR #325, opened 2018-11-13
Update settings, in particular version number

NotSpecial:
Furthermore, I now ensured that the setup always uses the correct version
number.

This required me to remove the 'hack' to always install the current folder
from `requirements.txt`. This was quite unusual anyways.

As a result tox etc. needed to be adjusted.

### DIFF
--- a/amivapi/settings.py
+++ b/amivapi/settings.py
@@ -13,7 +13,7 @@ from datetime import timedelta
 
 from passlib.context import CryptContext
 
-VERSION = '1.1.3'
+VERSION = '2.0.0'
 
 # Sentry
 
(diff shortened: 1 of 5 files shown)


amivapi/settings.py
```diff
@@ -165,26 +165,30 @@
...
     },
     'license': {
         'name': 'GNU Affero General Public License',
         'url': 'https://github.com/amiv-eth/amivapi/blob/master/LICENSE',
-    }
+    },
+}
+
+SWAGGER_LOGO = {
+    # TODO: Replace with AMIVAPI logo endpoint (or similar)
```
temparus: TODO should not be in code. Create an issues instead.

NotSpecial: I can remove the TODO from the comment, if you think thats more appropriate.

temparus: To keep track of issues and enhancements, the issues should be used to have a searchable collection of items in one place. Having this TODO in the code might produce inconsistent situations.

It might be possible that we decide to keep it like that. There may be a discussion in the said issue about that and the issue might be closed. The TODO still exists in the code, but there is no intention to do it anymore.

PR #326, opened 2018-11-13
Add schema definitions for internal fields

NotSpecial:
The API now adds definitions for all internal fields to the schema, i.e.
_etag and so on. Eve already did this for the _id field, so I extended it
(mainly such that the online documentation can include it nicely).


amivapi/utils.py
```diff
@@ -20,7 +21,7 @@
 from flask import g
 
 
-def token_urlsafe(nbytes=None):
+def token_urlsafe(nbytes=32):
```
NotSpecial: Note: This just brings our function up-to-date with the python implementation, which uses 32 bytes by default.

NotSpecial:
Note: @temparus double checked the code as well.

PR #328, opened 2018-11-23
Fix domain initialization

NotSpecial:
Eve does not allow schemas to contain fields such as `_id` or `_etag`.

However, we want to create them for better documentation.

This is no problem if we add them after Eve's check (which cannot be disabled :/),
but in the current implementation, this would all mutate the domain dict defined in settings.

Thus, if a second app were to be created in the same python interpreter, the domain would already be filled and Eve would complain.

Until now, this was only an issue with the tests, so I reset the domain after each test run.
However, due to a small mistake, the CLI would call `create_app` twice.

Normally, this should not be an issue. However, the situation explained above would cause Eve to complain and crash.

Now I decided to fix the problem more properly:

Instead of a fixed dict in `settings.py`, a fresh dict is now initialized in `create_app`.

### DIFF
--- a/amivapi/tests/test_media.py
+++ b/amivapi/tests/test_media.py
@@ -42,7 +42,6 @@ class MediaTest(WebTestNoAuth):
                 }
             }
         })
-        print('I WAS HEREEE', 'test' in self.app.config['DOMAIN'])
 
     def _post_file(self):
         """Post file. Use BytesIO to be able to set the filename."""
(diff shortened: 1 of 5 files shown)


amivapi/tests/test_media.py
```diff
@@ -32,8 +32,17 @@ class MediaTest(WebTestNoAuth):
...
+        self.app.register_resource('test', {
+            'resource_methods': ['POST', 'GET'],
+            'item_methods': ['GET', 'DELETE'],
+            'schema': {
+                'test_file': {
+                    'type': 'media'
+                }
+            }
+        })
+        print('I WAS HEREEE', 'test' in self.app.config['DOMAIN'])
```
temparus: Please remove this debug output.

NotSpecial: Hahahaha oops.

PR #342, opened 2019-02-05
Metadata summary for study documents

NotSpecial:
Currently, we have a lot of metadata for study docs, but no efficient way to search through it.
In order to provide users a menu to select professors, for example, an application would have to download *all* study documents to learn the possible professors.

On the other hand, this information is much needed, as most students will not know from the top of their head how exactly their professors name might be written or if their lecture can be found as 'Analysis' or 'Analysis I'

This defeats the purpose of filtering study documents with metadata.

Thus, this commit adds the possibility to request a server-side summary of metadata by appending `summary` to the query string, e.g.

```/studydocuments?summary```

The response then contains a new meta field `_summary`, which lists the available values for all meta-data fields along with how many documents are available for them.

```
_summary:
    professor:
        Prof. Dr. Awesome: 10
        Prof. Okay: 5
        ...
    lecture:
        ...
    ...
```

(For sanity reasons, the `author` field is not included in the summary, as most authors will only have a single document and thus the summary would be very large)

The summary is efficiently computed using the MongoDB aggregation framework.

### DIFF
--- a/amivapi/studydocs/model.py
+++ b/amivapi/studydocs/model.py
@@ -38,7 +38,36 @@ file uploaders have additional permissions:
 
 ## Summary
 
-TODO(Alex)
+For more efficient searching of study documents, a *summary* of available
+metadata fields can be requested with the `summary` keyword in the query string:
+
+```
+/studydocuments?summary
+/studydocuments?where={some filter}&summary
+```
+
+When a summary is requested, the response will contain a new field
+`_summary`, which contains additional information on metadata fields
+listing all distinct values for each field along with the number of documents
+available.
+
+```
+_summary:
+    professor:
+        Prof. Dr. Awesome: 10
+        Prof. Okay: 5
+        ...
+    lecture:
+        ...
+    ...
+```
+
+The summary is only computed for documents matching the current `where` query,
+e.g. when searching for ITET documents, only professors related to ITET
+documents will show up in the summary.
+
+Furthermore, the `author` metadata field is excluded from the summary, as there
+are only very few documents for most authors, thus not much to summarize.
 """)
 
 
@@ -114,14 +143,15 @@ studydocdomain = {
                 'empty': False,
                 'nullable': True,
                 'default': None,
-                'no_html': True
+                'no_html': True,
             },
             'department': {
                 'example': DEPARTMENT_LIST[0],
                 'type': 'string',
                 'nullable': True,
                 'default': None,
-                'allowed': DEPARTMENT_LIST
+                'allowed': DEPARTMENT_LIST,
+                'allow_summary': True,
             },
 
             'lecture': {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: amivapi/studydocs/model.py [hunk 1]
```diff
@@ -33,9 +33,51 @@
...
+    lecture:
+        ...
+    ...
+```
+
+The summary is only computed for documents matching the current `where` query,
+e.g. when searching for ITET documents, only professors related to ITET
+documents will show up in the summary.
+
+Furthermore, the `author` metadata field is excluded from the summary, as there
```
cburchert: I am against this. I think knowing whose studydoc you are searching is a common usecase. I remember we had a few very popular authors of studydocs, so that is definitely something to search for.
Ignore this if author means uploader, I don't exactly remember how that distinction was. Uploader doesn't matter, but author does.

amivapi/studydocs/summary.py
```diff
@@ -0,0 +1,71 @@
...
+        {'$group': {'_id': '$' + fieldname, '_count': {'$sum': 1}}},
+    ])
+    return {item['_id']: item['_count'] for item in aggregation
+            if item['_id'] is not None}
+
+
+def _get_lookup():
+    """Get the where clause lookup just like Eve does.
+
+    Unfortunately, Eve onlt parses the `where` at a very low level and does not
```
cburchert: only

amivapi/studydocs/summary.py
```diff
@@ -0,0 +1,71 @@
...
+    provide any methods to elegantly access it, so we have use the same
+    internal functions as Eve does.
+    (Recently, Eve has at least somewhat exposed the parsing, but this
+    code is part of an official release yet [1])
```
cburchert: is not?

amivapi/studydocs/model.py
NotSpecial: I am not sure. While there might be a handful of authors with multiple documents, the vast majority will probably just have a single document -- won't that just produce a huge summary with lots of values with only a single entry?

But you are right, maybe it is too early to optimize for this.

I can include the author without problems. Should it cause any problems in the future it can easily be disabled.

NotSpecial:
The reason we want this to be in the same resource is that it is closely related to the filters applied to studydocs and provides hints similar to HATEOAS.

In a way, the `_summary` tells you which filters are reasonable to apply next.

NotSpecial:
Further discussion with @cburchert:

It is a bit premature optimization to not just always include the summary. With the expected number of study docs (less than 1000), computing the summary should be nearly no overhead.

By always including it, we make it easier for clients to discover this information, otherwise someone might never know that this options exists.

cburchert:
Agreed to everything Alex just wrote. In addition for reference: If performance actually becomes a problem at any point, this should be easy to address. Either it can be solved on the database site with indices that pre-compute this aggregation or one could also simply cache the output of the _count_distinct function for a few minutes in memory.

PR #345, opened 2019-02-12
Media fix

NotSpecial:
I have noticed that even after #344, uploaded files would be corrupted, even if at least the API doesn't crash anymore.

After a bit of investigation I have discovered that the validator parses the file stream to determine width and height, but never resets the stream.

I have updated the media tests to now *always* ensure that files are uploaded correctly, and fixed the bug.

### DIFF
--- a/amivapi/validation.py
+++ b/amivapi/validation.py
@@ -310,7 +310,6 @@ class ValidatorAMIV(Validator):
                                "ratio. The accepted ratio is "
                                "%s:%s" % aspect_ratio)
 
-
     def _validate_session_younger_than(self, threshold_timedelta, field, _):
         """Validation of the used token for special fields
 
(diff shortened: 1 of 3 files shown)


amivapi/validation.py
```diff
@@ -292,8 +292,9 @@ def _validate_aspect_ratio(self, aspect_ratio, field, value):
         """
         width, height = aspect_ratio
         error = False
-        img = Image.open(value.stream)
-        print(img.size, aspect_ratio)
+        # Load file (and reset stream so it can be saved correctly afterwards)
+        img = Image.open(value)
```
NotSpecial: Note: `value` is a file object and can be directly read from, there is no need to access `value.stream` explicitly, thats why I have removed it.

PR #346, opened 2019-02-13
Use correct ratio (portrait) for event posters, update docs.

NotSpecial:
We specified the aspect ratio as height:width, but validate it width:height, which I now corrected.

While at it, I improved the respective docs a little.

### DIFF
--- a/amivapi/events/model.py
+++ b/amivapi/events/model.py
@@ -28,8 +28,8 @@ versions should be included.
 ## Images
 
 While the API generally accepts both `form-data` and JSON-input,
-JSON cannot be used to upload event images.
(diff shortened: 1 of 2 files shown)


amivapi/events/model.py
```diff
@@ -28,8 +28,8 @@
 ## Images
 
 While the API generally accepts both `form-data` and JSON-input,
-JSON cannot be used to upload event images.
-You must use [`multipart/form-data`][1] to be able to send files.
+images can only be send using [`multipart/form-data`][1]. There's a quick
```
temparus: wrong time form. Should be `sent`.

PR #356, opened 2019-02-27
change cents to rappen

hermannsblum:
The description text with 'cents' is more confusing than helpful to users.


NotSpecial:
Makes sense.

PR #363, opened 2019-03-30
Fix home link methods

### LINKED ISSUE #343: Unauthorized methods shown for some resources.
In our modified HATEOAS engine, we display all allowed methods.

For e.g. `/users`, without a token the response correctly contains only the `OPTIONS` method.

However, for the following resources, the API returns unauthorized methods:

- groupmemberships
- studydocumentss
- joboffers

In all cases, *only* `OPTIONS` should be returned, yet the API claims `OPTIONS`, `GET`, and `POST` are allowed. (A quick `GET` to studydocs without auth shows that this is in fact not allowed.)

There might be an issue how the methods are parsed for the home endpoint (`/`), I have to investigate this further.

NotSpecial:
For the home endpoint, some link methods were incorrect, i.e. the API would report that some methods are allowed for the current user (although they were not) and some of the reported methods would not be allowed.

This was the result of to unrelated problems:

1. Not allowed `HEAD`: Eve automatically allows `HEAD` requests if `GET` is possible. However, if `GET` is set to `public_methods`, `HEAD` *is not* automatically included, and has to be set explicitly. Unaware of this, the API would report that `HEAD` is allowed, while it was not. I have corrected the respective resource configs such that `HEAD` is allowed whenever `GET` is, as it is intended.

2. Allowed `GET` if it is not: Some resources using `AdminOnlyAuth` would not create a user lookup filter, but instead `abort`. The API had no way to check this properly, and would report that `GET` is possible (even if it would be aborted. I have changed the output of creating a lookup filter such that not allowing *any* lookup for a user can be properly returned. Now the API can correctly determine whether `GET` is possible or not.

Aside from that, I improved the tests for links on `/` to actually test all reported methods.

Closes #343.

### DIFF
--- a/amivapi/auth/auth.py
+++ b/amivapi/auth/auth.py
@@ -140,9 +140,10 @@ class AmivTokenAuth(BasicAuth):
 
         Returns:
             dict: The filter, will be combined with other filters in the hook.
-                Return None or empty dict if no filters should be applied.
+                Return empty dict if no filters should be applied.
+                Return None if no lookup should be possible at for the user.
         """
-        return None
+        return {}
 
 
 class AdminOnlyAuth(AmivTokenAuth):
@@ -153,7 +154,7 @@ class AdminOnlyAuth(AmivTokenAuth):
         Therefore no results should be given. To give a more precise error
         message, we abort. Otherwise normal users would just see an empty list.
         """
-        abort(403)
+        return None
 
 
 # Decorators that will only call a function if auth conditions are met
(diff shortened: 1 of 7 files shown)


amivapi/auth/auth.py
```diff
@@ -140,9 +140,10 @@ def create_user_lookup_filter(self, user_id):
 
         Returns:
             dict: The filter, will be combined with other filters in the hook.
-                Return None or empty dict if no filters should be applied.
+                Return empty dict if no filters should be applied.
+                Return None if no lookup should be possible at for the user.
```
cburchert: grammar

amivapi/tests/auth/test_link_methods.py
```diff
@@ -357,23 +360,35 @@ def get_user_methods(self, response):
             if links['href'] == 'users':
                 return links['methods']
 
+    def test_allowed(self, **args):
```
cburchert: This is now also a test function, I guess it should be a utility function instead.
Also you could specify the arguments.

PR #365, opened 2019-03-30
Add blacklist

I98:
Add new resource blacklist and an event-signup validator to check if a user
is on the blacklist. People can be on the blacklist for multiple
separate reasons, therefore blacklist is a new resource and not a single
field in the user resource.

### DIFF
--- a/amivapi/blacklist/model.py
+++ b/amivapi/blacklist/model.py
@@ -14,6 +14,17 @@ People normally get blacklisted if they don't appear to an event they signed up
 for, but other cases could be possible (Bad behaviour, etc). Once on the
 blacklist, they shouldn't be able to sign up for any event until they pay
 for that event or do something else the board decides (e.g. help at an event).
+
+An entry on the blacklist always has a reason and a start_date (normally the
+date of the event where the user didn't appear), most often also a price the
+user has to pay to be removed from the blacklist.
+
+One person can have multiple blacklist entries and old entries are not deleted,
+but marked with an end date.
+
+Only users with admin-rights for the blacklist can see all entries and create
+new ones/edit them. A single user only has the right to see his own blacklist
+entries, but he cannot edit them.
 """)
 
 blacklist = {
@@ -46,18 +57,24 @@ blacklist = {
                 'description': 'The reason for the blacklist entry',
                 'example': 'An Event xy nicht erschienen',
 
+                'empty': False,
                 'nullable': False,
                 'type': 'string',
                 'maxlength': 100,
                 'required': True
             },
             'price': {
-                'description': "The price",
+                'description': 'The price the user has to pay to be removed (in'
+                ' CHF) from the blacklist. Normally this is the event price, '
+                'but it can be zero if the user has to do something else, for '
+                'example help at an event. In the case of a free event, the '
+                'price should be 5.-',
                 'example': "5",
 
                 'type': 'integer',
                 'nullable': True,
-                'required': True,
+                'required': False,
+                'default': None,
             },
             'start_time': {
                 'description': 'The date on which the user was blacklisted',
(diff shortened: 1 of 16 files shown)


amivapi/blacklist/model.py
```diff
@@ -0,0 +1,81 @@
...
+"""
+from amivapi.blacklist.authorization import BlacklistAuth
+
+
+blacklist_description = ("""
+People normally get blacklisted if they don't appear to an event they signed up
+for, but other cases could be possible (Bad behaviour, etc). Once on the
+blacklist, they shouldn't be able to sign up for any event until they pay
+for that event or do something else the board decides (e.g. help at an event).
+""")
```
NotSpecial: The description is a bit short.

In particular, it does not cover:

1. That a user can be on the blacklist for multiple offences at once.
2. That the blacklist entry is defined by a start and end date
3. That users can only see their (own) blacklist entries, not others, and that only admins can set the dates.

Have a look at the other resource descriptions if you need inspiration :)

amivapi/blacklist/model.py
```diff
@@ -0,0 +1,81 @@
...
+                'nullable': False,
+                'required': True
+            },
+            'reason': {
+                'title': 'Reason',
+                'description': 'The reason for the blacklist entry',
+                'example': 'An Event xy nicht erschienen',
+
+                'nullable': False,
+                'type': 'string',
```
NotSpecial: You should also add the `empty: False` validation, to forbid empty strings.

amivapi/blacklist/model.py
```diff
@@ -0,0 +1,81 @@
...
+                'maxlength': 100,
+                'required': True
+            },
+            'price': {
+                'description': "The price",
+                'example': "5",
+
+                'type': 'integer',
+                'nullable': True,
+                'required': True,
```
NotSpecial: A field should not be both `required` and `nullable` -- if the fields can be just `None`, it is not required. Furthermore, if it is not required, it should include a default value.

amivapi/blacklist/model.py
```diff
@@ -0,0 +1,81 @@
...
+                'description': 'The date on which the user was blacklisted',
+                'example': '2018-10-11T00:00:00Z',
+
+                'type': 'datetime',
+                'nullable': False,
+                'required': True,
+            },
+            'end_time': {
+                'description':
+                'The date on which the user is deleted from the blacklist',
```
NotSpecial: Clarification: "The date on which the blacklist entry is deleted" -- the user could still be on the blacklist for other issues.

amivapi/events/validation.py
```diff
@@ -63,6 +63,21 @@ def _validate_json_event_field(self, enabled, field, value):
...
                 self._error(field, error.message)
 
+    def _validate_not_blacklisted(self, enabled, field, user_id):
+        """Validate if the user is not on the blacklist.
+
+        The rule's arguments are validated against this schema:
+        {'type': 'boolean'}
+        """
+        if enabled:
+            count = current_app.data.driver.db['blacklist'].count({
```
NotSpecial: It seems that the `count` operation [is deprecated](https://docs.mongodb.com/manual/reference/method/cursor.count/) and `count_documents` [should be used instead](http://api.mongodb.com/python/current/api/pymongo/collection.html?highlight=count_documents#pymongo.collection.Collection.count_documents). It seems to have the same api as `count`, so this should be an easy change.

amivapi/blacklist/model.py
```diff
@@ -0,0 +1,81 @@
...
+                'title': 'Reason',
+                'description': 'The reason for the blacklist entry',
+                'example': 'An Event xy nicht erschienen',
+
+                'nullable': False,
+                'type': 'string',
+                'maxlength': 100,
+                'required': True
+            },
+            'price': {
```
NotSpecial: I am not sure about this price field. Is this always the price of a corresponding event? Is the price in rappen, like for events? I feel this field requires more explanation.

PR #371, opened 2019-03-30
Add event moderator

Mat-pa:


amivapi/tests/events/test_auth.py
```diff
@@ -241,3 +241,49 @@ def test_checkin_admin_permissions(self):
...
+        # user
+        self.assertEqual(self.api.get("/eventsignups?where={\"event\":\""
+                                      + str(ev['_id']) + "\"}",
+                                      token=user_token, status_code=200).
+                         json['_meta']['total'], 1)
+        # moderator
+        self.assertEqual(self.api.get("/eventsignups?where={\"event\":\""
+                                      + str(ev['_id']) + "\"}",
+                                      token=moderator_token, status_code=200).
+                         json['_meta']['total'], 2)
```
NotSpecial: Could you add a test to ensure that the moderator cannot create or delete signups (and only see them)?

NotSpecial:
I needed to make some small adjustments, which are now part of #377.

PR #384, opened 2019-04-02
Improve mail function

NotSpecial:
All API mails are sent from the same address, thus the `mail` util
does not require a sender mail anymore, and instead always uses
the API mail from the config.

Additionally, the mail function now adds an configurable prefix
to the mail subject, e.g. "[AMIV]", to increase consistency and
allow users to spot API mails quickly.

### DIFF
--- a/amivapi/settings.py
+++ b/amivapi/settings.py
@@ -65,6 +65,7 @@ REMOTE_MAILING_LIST_DIR = './'  # Use home directory on remote by default
 
 # SMTP server defaults
 API_MAIL = 'api@amiv.ethz.ch'
+API_MAIL_SUBJECT_PREFIX = "[AMIV]"
 SMTP_HOST = 'localhost'
 SMTP_PORT = 587
 SMTP_TIMEOUT = 10
--- a/amivapi/settings.py
+++ b/amivapi/settings.py
@@ -65,7 +65,7 @@ REMOTE_MAILING_LIST_DIR = './'  # Use home directory on remote by default
 
 # SMTP server defaults
 API_MAIL = 'api@amiv.ethz.ch'
-API_MAIL_SUBJECT_PREFIX = "[AMIV]"
+API_MAIL_SUBJECT = "[AMIV] {subject}"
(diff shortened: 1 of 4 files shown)


amivapi/settings.py
```diff
@@ -65,6 +65,7 @@
 
 # SMTP server defaults
 API_MAIL = 'api@amiv.ethz.ch'
```
hermannsblum: Do we want to change this default to `info@amiv.ch`? We had users actually emailing to the api email when something with their eventsignups went wrong, and the API admins are definitely not the right people to handle this. It is confusing to users that you can answer and then the email does not go to people responsible for the event, but the developers of the API.

NotSpecial: Good point. But this can be done in the settings as well, so I think changing the default is not a priority. For that I guess we first need a discussion which email is the appropriate one.

For that, we might include a 'reply-to' field.

amivapi/utils.py
```diff
@@ -74,15 +74,21 @@ def get_id(item):
...
+
+
     Args:
-        from(string): From address
         to(list of strings): List of recipient addresses
         subject(string): Subject string
         text(string): Mail content
     """
+    sender = app.config['API_MAIL']
+    subject = " ".join((app.config['API_MAIL_SUBJECT_PREFIX'], subject))
```
hermannsblum: Instead of a configurable prefix, I think a template string is more intuitive when looking into the config.

amivapi/settings.py
hermannsblum: ok my comment jsut became the review summary, anyways: 

Yes, `reply-to` would actually be what we want. Please add it to the config here.

amivapi/utils.py
NotSpecial: Fair enough

NotSpecial:
I cannot add a reply-to field in this PR. While I generally agree, I think we first need a discussing on which reply-to addresses to use where. I think this is orthogonal to this PR.

PR #392, opened 2019-04-27
Don't install dependencies in editable mode

cburchert:
We added this to make it easier to insert print statements into our dependencies or do other code changes for testing. We don't want this in production and it is simple to readd it for debugging if needed.


NotSpecial:
Makes sense, approved :)

NotSpecial:
I have included this change in the general dependency update in #396.

PR #395, opened 2019-05-30
Allow proper caching with `If-Modified-Since`.

### LINKED ISSUE #389: Add separate Cache-Control header for media endpoint
Media files are currently not cached. It would make sense to allow caching of those files as they take some time to load whenever one navigates on the AMIV website and they do not change that often.

I think that it would be reasonable to allow caching for one hour just for the `/media` endpoint.

NotSpecial:
There is a subtle difference between `no-store` and `no-cache`.

`no-store`: Never return from cache, always request new. (Especially problematic for images, which are always re-loaded.) **This is what we currently have**.

`no-cache`: Don't return from cache by default, but first re-validate with the server, in our case using the `If-Modified-Since` header, which browsers use automatically, as eve includes `Last-Modified` headers in all responses. **This is what we actually want.**

Thus, with this small change, the api should cache in the way we want. Stale results from cache will never be returned without a request, and data is only transmitted if anything has changed.

More info: https://www.keycdn.com/blog/http-cache-headers

Note: Having both `no-cache` and `must-revalidate` doesn't actually do anything, as each on it's own would be enough for what we want, but for compatibility its probably best to keep both.

Closes #389

### DIFF
--- a/amivapi/settings.py
+++ b/amivapi/settings.py
@@ -41,7 +41,7 @@ X_DOMAINS = '*'
 X_HEADERS = ['Authorization', 'Content-Type', 'Cache-Control',
              'If-Match', 'If-None-Match', 'If-Modified-Since']
 MONGO_QUERY_BLACKLIST = ['$where']  # default blacklists where and regex queries
-CACHE_CONTROL = 'no-store, must-revalidate'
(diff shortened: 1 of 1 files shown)


temparus:
The tests seem to fail. Could you have look on that?

PR #396, opened 2019-06-02
Update requirements and test setup.

### LINKED ISSUE #394: Pagination info totals wrong in GET request to /groupmemberships with embedded groups
As pagination is enabled for the AMIV API and you request multiple data items (e.g. /users or /groupmemberships), the API supplies the items under the _items key. As at maximum 25 items are returned, some pagination info is supplied under the _meta key. In the _meta key, the current page (page), the maximum number of items per page (max_results), and the total number of results for the request (total) are delivered.

Currently, when GETting the following request:
`api-dev.amiv.ethz.ch/groupmemberships?where={"user":"5bc130c27fbf9a0001a284a5"}&embedded={"group":1}`
The _items list has 2 objects in it, but the _meta->total value shows only a 1.
Not embedding the group results in the correct _meta->total value.

We noticed this error because it triggered some internal AMIV API response sanity checks within amiv-checkin. #youcanneverdoenougherrorchecking

NotSpecial:
For some reason, the generic alpine image would not correctly update
the repositories and would not find mongodb.

With a fixed version, this issue does not occur anymore.

Additionally, important changes due to version update:

- Pytz: Dependency removed, as datetime
  can handle utc on its own.

- Cerberus: `valueschema` renamed to `valuesrules`.

- MongoDB: `insert` is replaced by `insert_one` and `insert_many`.
  As a consequence, ObjectIDs are retrieved
  a bit differently.

- PassLib: `encrypt` renamed to `hash`.
  `vary_round` is deprecated, has it
  "provided little security and was not worth
  the additional complexity" [1].

[1]: https://passlib.readthedocs.io/en/stable/history/1.7.html?highlight=vary_rounds#crypt-context-api-deprecations

Closes #394 (new eve version fixes this bug)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,7 +4,6 @@ Dockerfile.test
 run_tests.sh
 
 # Byte-compiled / optimized / DLL files
-__pycache__/
 **/__pycache__/
 *.py[cod]
 
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,6 +5,7 @@ run_tests.sh
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
+**/__pycache__/
 *.py[cod]
 
 # Editor backups
--- a/amivapi/events/validation.py
+++ b/amivapi/events/validation.py
@@ -9,7 +9,6 @@ import json
 
 from flask import current_app, g, request
 from jsonschema import Draft4Validator, SchemaError
-import pytz
 
 
 class EventValidator(object):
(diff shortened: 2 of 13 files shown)


amivapi/events/validation.py
```diff
@@ -114,11 +113,11 @@ def _validate_signup_requirements(self, signup_possible, field, event_id):
 
             # The event has signup, check if it is open
             if not g.get('resource_admin'):
-                now = datetime.now(pytz.utc)
-                if now < event['time_register_start']:
+                now = datetime.utcnow()
```
NotSpecial: Note: This creates a utc time, but without any timezone info.
Therefore we need to use `.replace(tzinfo=None)` below, as the event times are
utc as well, but *with* timezoneinfo.

.dockerignore
```diff
@@ -5,6 +5,7 @@ run_tests.sh
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
```
cburchert: I think this is now part of the next line as well, as ** can also match zero directories.

amivapi/events/validation.py
cburchert: I think the replace doesn't do anything. utcnow() returns already a naive datetime object with tzinfo None, why would replacing this None with a None do anything? If the db returns things with timezones, then the old code was the correct way to get the current UTC.

cburchert: Ok nevermind the replace is on the event time not on the `now` object.

PR #397, opened 2019-06-03
Add ratings for study documents.

### LINKED ISSUE #89: Rating for /studydocuments
There's a need for an average rating parameter for the studydocuments. (Average rating/ Number of ratings)

NotSpecial:
Add a rating for study documents, via a new endpoint `studydocumentratings` to rate, and a field `rating` to display the result.

We do not simply return the average vote, which can be quite far off, since we don't have many votes in general.

Instead, we return the lower bound of a statistical confidence interval.

On the one hand, it is useful, but actually I thought it's a neat approach, and since it is not much more complicated than an average rating I went for it^^
(The actual formula is a bit more complex, but thats it)

[More info :)](http://www.evanmiller.org/how-not-to-sort-by-average-rating.html)

Closes #89

### DIFF
--- a/amivapi/studydocs/model.py
+++ b/amivapi/studydocs/model.py
@@ -36,6 +36,12 @@ file uploaders have additional permissions:
 
 <br />
 
+## Ratings
+
+TODO!
+
+<br />
+
 ## Summary
 
 For more efficient searching of study documents, a *summary* of available
@@ -196,7 +202,73 @@ studydocdomain = {
                 'description': 'The year in which the course *was taken*, '
                                'to separate older from newer files.',
                 'allow_summary': True,
-            }
+            },
+
+            'rating': {
+                'title': 'Study Document Rating',
+                'description': 'The study document rating as a fraction of '
+                               'upvotes divided by all votes. Computed using '
+                               'a confidence interval. Null if no votes have '
+                               'been cast.',
+                'example': '0.9',
+
+                'type': 'float',
+                'readonly': True,
+            },
+        },
+    },
+
+    'studydocratings': {
+        'resource_title': "Study Document Rating",
+        'item_title': "Study Document Rating",
+
+        'description': "Rating for Study documents (TODO)",
+
+        'resource_methods': ['GET', 'POST'],
+        'item_methods': ['GET', 'PATCH', 'DELETE'],
+
+         # TODO
+        'authentication': StudydocsAuth,
+
+        'schema': {
+            'user': {
+                'description': 'The rating user.',
+                'example': '679ff66720812cdc2da4fb4a',
+
+                'type': 'objectid',
+                'data_relation': {
+                    'resource': 'users',
+                    'embeddable': True,
+                    'cascade_delete': True,
+                },
+                'not_patchable': True,
+                'required': True,
+                'unique_combination': ['studydoc'],
+            },
+
+            'studydoc': {
+                'title': 'Study Document',
+                'description': 'The rated study doc.',
+                'example': '10d8e50e303049ecb856ae9b',
+
+                'data_relation': {
+                    'resource': 'studydocuments',
+                    'embeddable': True,
+                    'cascade_delete': True,
+                },
+                'not_patchable': True,
+                'required': True,
+                'type': 'objectid',
+            },
+            'rating': {
+                'description': 'The given rating, can be an up- or downvote.',
+                'example': 'up',
+
+                'type': 'string',
+                'allowed': ['up', 'down'],
+                'required': True
+            },
         },
     }
+
 }
--- a/amivapi/studydocs/model.py
+++ b/amivapi/studydocs/model.py
@@ -4,9 +4,10 @@
 #          you to buy us beer if we meet and you like the software.
 """Model for studydocuments."""
 
-from amivapi.settings import DEPARTMENT_LIST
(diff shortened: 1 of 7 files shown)


amivapi/studydocs/model.py
```diff
@@ -196,7 +233,73 @@ def _validate_allow_summary(self, *args, **kwargs):
                 'description': 'The year in which the course *was taken*, '
                                'to separate older from newer files.',
                 'allow_summary': True,
-            }
+            },
+
```
cburchert: Nit: this empty line is inconsequenctial with the rest of the schema.

amivapi/studydocs/rating.py
```diff
@@ -0,0 +1,81 @@
...
+    (As there are only a few votes, we do not want to be overly cautious).
+
+    [1]: http://www.evanmiller.org/how-not-to-sort-by-average-rating.html
+    """
+    total = upvotes + downvotes
+    if not total:
+        return
+
+    p = upvotes / total
+    bound = ((p + (z**2)/(2*total) -
```
cburchert: not sure if python optimizes this somehow, but often z*z is vastly faster than z**2, as it uses a simple multiplication instead of a long algorithm.

amivapi/studydocs/rating.py
```diff
@@ -0,0 +1,81 @@
...
+    upvotes = ratings.count_documents({'rating': 'up', **lookup})
+    downvotes = ratings.count_documents({'rating': 'down', **lookup})
+
+    # Compute rating and write to database
+    rating = compute_rating(upvotes, downvotes)
+    docs.update_one({'_id': studydoc_id}, {'$set': {'rating': rating}})
+
+
+def init_rating(items):
+    """On creating of a study-document, set the rating to None."""
```
cburchert: On creation

amivapi/studydocs/rating.py
```diff
@@ -0,0 +1,81 @@
...
+    for item in items:
+        item['rating'] = None
```
cburchert: How about 0? That would mathematically make sense and be the result of upvoting and then deleting your upvote as well. By adding None here we get just another case to handle for every user interface. Treating 0 as unrated should be fine anyway.

amivapi/tests/studydocs/test_rating.py
```diff
@@ -0,0 +1,229 @@
...
+        """Ensure that sorting by rating works."""
+        ids = [24 * '0', 24 * '1', 24 * '2']
+        self.load_fixture({
+            'users': [{}],
+            'studydocuments': [
+                {'_id': ids[0]}, {'_id': ids[1]}, {'_id': ids[2]}
+            ],
+            'studydocumentratings': [
+                # Rate such that:
+                # 0: No rating, 1: Low rating, 2: High rating
```
cburchert: This is a slightly weird ordering... But that is just what the math works out to. We could in theory give unrated things a slightly positive epsilon rating to put them above things only voted negative.

amivapi/studydocs/rating.py
cburchert: Ah I was wrong about the upvoting and then deleting. That gets reset to None.

NotSpecial:
Thanks for the review, and you raise a valid point with the ordering. However, any default value also seems a bit weird to me, in that case there is no clear way to distinguish a 'slightly' bad studydoc from an unrated one.

Aside from those concerns, where would you put this threshold? I.e. at which rating would you place 'bad' documents under unrated ones?

cburchert:
Not sure what you mean by 'slightly bad' studydoc. You would need a huge number of votes to get to an epsilon lower bound, right?

Basically I would just return something like 1e-6 (large enough that it doesn't disappear in floating accuracy) from compute_rating() if there are no votes at all. As all cases where there are 0 upvotes, but >0 downvotes result in a value of 0 already, the ranking would always be:

1. stuff with at least 1 upvote.
2. stuff with 0 votes.
3. stuff with 0 upvotes, >0 downvotes.

There is of course some uncertainty for things that have 0 upvotes and 1 downvote and thus drop below unrated things, but I would guess this is still better than having the None value.

Another more complex suggestion would be to order things with the same lower confidence bound by their upper confidence bound. That would have a similar effect, but would also rank the stuff with only downvotes by ascending downvotes.

amivapi/studydocs/rating.py
```diff
@@ -0,0 +1,81 @@
...
+    if not total:
+        return
+
+    p = upvotes / total
+    bound = ((p + (z**2)/(2*total) -
+              z * sqrt((p * (1-p) + (z**2)/(4*total)) / total)) /
+             (1 + (z**2)/total))
+
+    # Ensure that the bound is not below 0
+    return max(bound, 0)
```
cburchert: Can `bound` mathematically even be lower than 0?

amivapi/studydocs/rating.py
NotSpecial: Not sure either, but I do not think it matters here. The algorithm runs once per vote, so the performance difference will be negligible.

NotSpecial:
I have thought about this issue again. I think it might be best to just return more to the client, i.e. return both the upvotes, downvotes as well as the rating. This way, the client can still sort by rating, but show upvotes and downvotes, which might be more informative to the common user than the rating.

PR #398, opened 2019-09-16
Fix ldap membership detection (students >2019).

NotSpecial:
As of fall 2019, the LDAP data for students has changed.
The general string from which we detect membership has moved from the `departmentNumber` field to the (new) `description` field.
Unfortunately, the info for all previous students is still in the
`departmentNumber` field, thus we now check which of both fields is in use before
parsing.


NotSpecial:
I have tested the integration with the nethz shortnames `adietmue`, `lutzsa` and `aheidemue` (a shortname of a student that was previously *not* imported, provided by @temparus).

PR #401, opened 2019-10-05
Fix crashes when trying to send blacklist emails

### LINKED ISSUE #391: Blacklist `end_time` might be None
The value for `end_time` of the resource `blacklist` might be `None`.

See the error report [AMIV-API-1C](https://sentry.amiv.ethz.ch/share/issue/4220f482564a43acb39195e485665b19/).

cburchert:
Fix #391.
We did not correctly handle that the end_time of an event is nullable. To fix:
1. Patching a blacklist entry to remove the end_time now doesn't schedule an email about the entry timing out anymore.
2. When the scheduler runs an event to send an email and the end_time was deleted in the meantime we return instead of crashing.


amivapi/blacklist/emails.py
```diff
@@ -67,8 +69,9 @@ def notify_patch_blacklist(new, old):
     """Send an email to a user if one of his entries was updated."""
     # Checks if the particular update resolved the blacklist entry or just
     # fixes an error, for example changed the reason or price. An entry is
-    # resolved when the end_time is before now.
-    if 'end_time' not in new:
+    # resolved when the end_time is before now. The end_time might also
+    # have been removed, in which case we don't schedule an email either.
+    if 'end_time' not in new or new['end_time'] is None:
```
NotSpecial: Could technically be reduced to `if new.get('end_time') is None`, but I think this might be more easily understandable, thus now change required.

amivapi/blacklist/emails.py
```diff
@@ -26,19 +26,21 @@ def _get_email(item):
...
-    if _item['end_time'].replace(tzinfo=None) == item['end_time']:
-        email = _get_email(_item)
-        fields = {'reason': _item['reason']}
-        mail(email,
-             'Your blacklist entry has been removed!',
-             current_app.config['BLACKLIST_REMOVED'].format(**fields))
+    if _item.get('end_time') is None:
+        return  # Entry was patched to last indefinitely, so no mail to send.
+    if _item['end_time'].replace(tzinfo=None) != item['end_time']:
+        return  # Entry was edited, so this is outdated.
```
NotSpecial: Very nicely understandable!

PR #413, opened 2019-11-26
Fix FCFS POST issue for admins (fix #381)

### LINKED ISSUE #381: For FCFS, admins can PATCH accepted, but not POST.
Admins can manage singups manually, which is both intended and required, see #362.

However, when FCFS is on, this only works for `PATCH`. On the other hand, for `POST`, any data sent for the field `accepted` is simply overridden by the API, which should not happen.

temparus:
Fixes the accepted signup POST issue when the action is performed by an admin.

### DIFF
--- a/amivapi/events/queue.py
+++ b/amivapi/events/queue.py
@@ -4,7 +4,7 @@
 #          you to buy us beer if we meet and you like the software.
 """Logic to implement different signup queues."""
 
-from flask import current_app
+from flask import current_app, g
 from pymongo import ASCENDING
 
 from amivapi.events.emails import notify_signup_accepted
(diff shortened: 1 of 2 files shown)


amivapi/events/queue.py
```diff
@@ -65,7 +65,7 @@ def update_waiting_list(event_id):
 def add_accepted_before_insert(signups):
     """Add the accepted field before inserting signups."""
     for signup in signups:
-        signup['accepted'] = False
+        signup['accepted'] = g.resource_admin and signup.get('accepted', False)
```
NotSpecial: I think this line needs a comment above, e.g.

```python
# admins may provide a value for `accepted`.
#if not provided or not admin set it to false`.
```

I'm afraid without a comment, nobody will understand what this line is doing.

Maybe it would be useful to just put this in the docstring.

amivapi/tests/events/test_queue.py
```diff
@@ -55,6 +55,15 @@ def test_fcfs_users_get_auto_accepted(self):
                                     status_code=200).json
         self.assertTrue(user2_signup['accepted'])
 
+        # post accepted signup as admin
```
NotSpecial: It is not clear that this is tested as an admin, I would maybe not use the `webtestnoauth` here.

Additionally, please also test the other direction, i.e. that a non-admin *cannot* set `accepted: True`.

PR #416, opened 2019-12-01
Fix accepted mail sending when present in eventsignups POST

### LINKED ISSUE #415: No email sent when eventsignup POST contains accepted:true
An admin can always provide the `accepted` field for an `eventsignup` set to `true`.

If one does so with a POST request, no email is sent to the person who was signed up by the admin.

Expected behavior would be that the signed up person also gets the email that they were signed up. Otherwise, the user might not know that he/she is on the participants list of that event.

temparus:
Ensures that an accepted email is sent to the user when an admin created an event signup directly with `accepted = true`.

Fixes #415.

### DIFF
--- a/amivapi/tests/events/test_emails.py
+++ b/amivapi/tests/events/test_emails.py
@@ -4,6 +4,7 @@
 #          you to buy us beer if we meet and you like the software.
 """Test email confirmation system for external event signups."""
 
+import datetime
 import re
 
 from amivapi.tests.utils import WebTestNoAuth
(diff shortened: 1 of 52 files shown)


NotSpecial:
@temparus  any updates on this one?

amivapi/tests/events/test_emails.py
```diff
@@ -229,3 +229,39 @@ def test_no_nones_in_emails(self):
...
+
+        self.api.post('/eventsignups', data={
+            'user': str(user['_id']),
+            'event': str(event['_id']),
+            'accepted': True
+        }, token=self.get_root_token(), status_code=201).json
+
+        mail = self.app.test_mails[0]
+        for field in mail.values():
+            self.assertTrue('None' not in field)
```
NotSpecial, line 249: Does it have to be `'None'` and not just `None`? (Not sure, I am not familiar with the mail tests, sorry).

NotSpecial (commented):
Thanks for the updates. I have just one comment on the test because I am unfamiliar with the mail-related tests.

amivapi/tests/events/test_emails.py
temparus: We have to check for the string `'None'` to make suer that neither the subject nor the email body contain that phrase as it would indicate that some placeholder was not set properly.

NotSpecial: And subjet and body are guaranteed to be strings, they can never be none?

NotSpecial: I did not write the mail test so I am not familiar with this. Could you double check this please?

In particular, can you check why it's `'None'` (string) not not just `None`? If we know the reason I am happy to acccapt it, but like this its a bit shady.

amivapi/tests/events/test_emails.py
```diff
@@ -229,3 +229,39 @@ def test_no_nones_in_emails(self):
...
+
+        # accepted email and confirmation email should be sent.
+        self.assertIs(len(self.app.test_mails), 3)
+        for mail in self.app.test_mails[1:]:
+            for field in mail.values():
+                self.assertTrue('None' not in field)
+
+            self.assertTrue(
+                ('accepted' in mail['subject']) or
+                ('confirm_email' in mail['text']))
```
NotSpecial, line 267: Arg, a bit the same issue as above. These assertions seem super fishy. I guess it's because of the way the test is set up? Maybe we can add a comment to explain why asserting this means that everything works?

NotSpecial (changes requested):

amivapi/tests/events/test_emails.py
temparus: Well, the first test case with the None values in the templated string fields is pretty obvious to me. Nevertheless I just added a comment there which additionally explains it in English.

So to explain it again with an example:
A confirmation email for the event signup of event XYZ should be sent. This is a prepared message containing placeholders. So the event name is a placeholder which gets replaced by 'XYZ'. In case the event name is None for some weird reason, the event name then listed in the message is 'None'.

suisseWalter: this looks like a satisfactory explanation to me.

suisseWalter (approved):
This testcase looks good to me. and as it just adds a testcase for something that hasn't been tested in the past it should be fine.

PR #423, opened 2021-01-27
Studydocs: Always require type and add options.

### LINKED ISSUE #422: Make the type of a document a necessary field for the Study Documents
To prepare for some QoL improvements in the frontend (displaying all documents sorted by their type from the get go) we require all data to contain a valid type field. Thus the API should implement this requirement in the backend.

NotSpecial:
We do not wish to allow studydocs without type anymore, so the `type` field is not not nullable and required.

Furthermore, we wish to differentiate between spring and autumn exams for better filtering, so the previous exam type is replaced.

Closes #422

### DIFF
--- a/amivapi/studydocs/model.py
+++ b/amivapi/studydocs/model.py
@@ -182,9 +182,9 @@ studydocdomain = {
             'type': {
                 'example': 'cheat sheets',
                 'type': 'string',
-                'nullable': True,
-                'default': None,
-                'allowed': ['exams', 'oral exams',
+                'required': True,
+                'nullable': False,
+                'allowed': ['spring exams', 'autumn exams', 'oral exams',
                             'cheat sheets', 'study guides',
                             'lecture documents', 'exercises'],
                 'allow_summary': True,


temparus:
The tests for `amivapi/tests/studydocs/test_studydocs_auth.py` and `amivapi/tests/studydocs/test_summary.py` are failing.

NotSpecial:
Ah, because type is now required, and of course the old tests do not specify the type.

NotSpecial:
If anybody has time, this should be an easy fix. I'll do it as soon as I can, but not sure when that will be :/

temparus:
@NotSpecial The change of the type options for study documents do not come with any data migrations, so the existing study documents will still come with the type `exam`. This case is not covered by the tests as the API does not accept an item with type `exams` anymore.

Do you have any idea how the API handles this situation?

temparus:
I just saw the use of `self.load_fixture(...)` in other tests. We could add some additional tests to cover that scenario by adding the entities with `load_fixture` instead of the regular POST endpoint of the API.

NotSpecial:
The API will enforce the new type for all new requests. Existing data will be returned as-is. We would need to migrate it ourselves.

temparus:
Migration of existing data is difficult because we can only partially automate the process by looking up lectures in the course catalogue of ETH and match those lectures which still exist with the same name.
If we keep the old entries as is, we might need to have additional filter options in addition to the document types officially supported by the API in order to be able to find all documents by filtering. This is also not ideal.

NotSpecial:
I'm not sure what to tell you 😅 

Data migration is always a difficult task, that's why changes to the data schema are only done when really necessary.

So considering that, is the split into autumn/spring exams really that necessary?
We could easily revert this part of the PR and go back to just `exam`.

I was just implementing the suggestion in #422,  but now that I see it again I am not sure whether it's really that helpful to split the exam types up.

What's your opinion? Should we just keep `exam` as before? I mean usually, the exam is in the same semester as the lecture, so looking for lectures in HS with an exam is the same as looking for lectures with an autumn exam.

temparus:
I think that having the document type `exam` would be sufficient.
It can happen that a lecture gets moved from the autumn semester to the spring semester or vice versa. When this happens, I would want to be able to get all exams for that lecture from the previous years nonetheless. So in this case, splitting the type would rather introduce more confusion.

I will check with the current board members how we should proceed and then get back to you.

d-portenier:
I haven't talked to the rest of the Board, but I think the separation into spring/ autumn exams is more hindering than helpful, as if I'm looking for old exams, I also want to find the ones in the off-season.

Additionally, people often mix things up. In the past, I downloaded old exams with 'spring20' in the doc-title, but the exam date was February 20 and should therefore have been 'autumn19'.

NotSpecial:
In that case, we should probably revert this part of the change and go back to a single `exam` type.

temparus (approved):

temparus:
Please note that we might still have existing documents where this information is not set, so these documents should be migrated after this change has been deployed.

amivapi/tests/studydocs/test_summary.py
```diff
@@ -57,6 +64,11 @@ def test_filtered_summary(self):
 
         assert '_summary' in response
         assert response['_summary'] == {
+            'type': {
+                # The document with title `third` is ignored
```
suisseWalter, line 68: what does this coment say? 
is it actually the case that calling a study doc "third" will not work?

suisseWalter (changes requested):
overall I like the idea. 
But I'm concerned about migration. 
also, because we will be migrating to community solutions in the future it might not make sense to do a lot of work on the migration, if that requires a lot of work.

amivapi/tests/studydocs/test_summary.py
temparus: The `where` clause used for the API request only matches documents where the title is `first` or `second`, so the document with the title `third` will not match the where condition.
Therefore the document is filtered out and should also not be counted in the summary part of the response.

temparus:
> overall I like the idea. But I'm concerned about migration. also, because we will be migrating to community solutions in the future it might not make sense to do a lot of work on the migration, if that requires a lot of work.

The existing documents must be migrated to the new platform, so if there is more structured data available to perform this migration as automated as possible the better it is.
This change only applies to newly uploaded or edited documents, but those documents are less likely to require manual review for migration.

suisseWalter:
ok if there is no need to migrate old documents then I'm happy with this PR.

suisseWalter (approved):

PR #425, opened 2021-03-16
Add setting 'LDAP_HOSTS'

### LINKED ISSUE #424: Make LDAP Server configurable
We should make the LDAP servers configurable so that we can remove bad servers from the list.

https://github.com/amiv-eth/amivapi/blob/427334aa745db40aac71b275a76c207a5e6e3a7d/amivapi/ldap.py#L63

temparus:
Resolves #424.


michbaum (approved):

temparus:
Related issue #424 was closed.

NotSpecial:
I'm not against this change, but why is it necessary?

In #424 , the last state is that the issue was a deleted LDAP account, not any issue with the hosts. Did this change?

temparus:
Recently, we had to change the LDAP Server addresses as the ones we used before have been shut down.
Unfortunately, the people did not find where to change that setting as it is not intuitive to find it hardcoded in a separate package.
As we only use that package in this project, we decided to add that configuration option to make it more intuitive for future IT people to change those URLs.

NotSpecial:
Thanks for clarifying, that makes sense!

So just to be clear, the server addresses have permanently changed?
(I am thinking of whether the ldap library should be updated, too)

temparus:
Yes, this change is permanent.
_(But we thought more about removing the default LDAP servers from the LDAP library than updating them as the amivapi is the only application currently using it)._

PR #426, opened 2021-03-16
Remove unnecessary package 'python-dev'

temparus:
The package `python-dev` was decomissioned with Python 2.
We do not need anything from this package so we should just remove it so that the docker build succeeds again.


cburchert (approved):
Thanks for the fix! This seems straightforward, so if the tests work I suppose you can merge.

NotSpecial:
Already resolved in #427.

PR #429, opened 2021-03-17
Allow login if LDAP auth works, but sync fails.

NotSpecial:
The LDAP sync requires additional credentials.
If there are problems with these, sync can fail
(even if auth works).

To reduce the damage, this change allows in-db users to still log in,
even if sync failed.
(As long as they can be authenticated).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ eve==1.1.5
 git+https://github.com/hermannsblum/eve-swagger.git@5c9cf066ffa72712f9018a3eb0c5142f7733d8cc#egg=Eve_Swagger
 # "nethz" must be installed in editable mode, otherwise some certs are not found
 # Wontfix: With the upcoming migration, this library will not be needed anymore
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -2,7 +2,7 @@ eve==1.1.5
 git+https://github.com/hermannsblum/eve-swagger.git@5c9cf066ffa72712f9018a3eb0c5142f7733d8cc#egg=Eve_Swagger
 # "nethz" must be installed in editable mode, otherwise some certs are not found
 # Wontfix: With the upcoming migration, this library will not be needed anymore
--e git+https://github.com/NotSpecial/nethz.git@1d3004081c3618f1f41463476a847b0bddd6d91a#egg=nethz
+-e git+https://github.com/NotSpecial/nethz.git@fcd5ced2dd365f237047748abfedb9c35a468393#egg=nethz
```
temparus, line 5: Shoudn't we use the fork in the amiv-eth group here?

temparus: We can also do this in another MR actually.

temparus (approved):
Great work! Thanks.

PR #439, opened 2022-05-07
Feature/reply to

emorbach:


amivapi/tests/events/test_emails.py
```diff
@@ -229,3 +229,29 @@ def test_no_nones_in_emails(self):
         mail = self.app.test_mails[1]
         for field in mail.values():
             self.assertTrue('None' not in field)
+
+    def test_reply_to_emails(self):
+        event = self.new_object('events', spots=100, selection_strategy='fcfs')
```
temparus, line 234:
```suggestion
        # Check that there is no `reply-to` header in the email if no moderator is set.
        event = self.new_object('events', spots=100, selection_strategy='fcfs')
```

amivapi/tests/events/test_emails.py
```diff
@@ -229,3 +229,29 @@ def test_no_nones_in_emails(self):
...
+        signup = self.api.post('/eventsignups', data={
+            'user': str(user['_id']),
+            'event': str(event['_id'])
+        }, status_code=201).json
+
+        mail = self.app.test_mails[0]
+
+        self.assertTrue('reply-to' not in mail)
+
+        user_moderator = self.new_object('users', email='xyz@gmail.com')
```
temparus, line 246:
```suggestion
        # Check that the `reply-to` header is present in the email if the moderator is set.
        user_moderator = self.new_object('users', email='xyz@gmail.com')
```

amivapi/tests/events/test_emails.py
```diff
@@ -229,3 +229,29 @@ def test_no_nones_in_emails(self):
...
+
+        signup = self.api.post('/eventsignups', data={
+            'user': str(user['_id']),
+            'event': str(event2['_id'])
+        }, status_code=201).json
+
+        mail = self.app.test_mails[1]
+        self.assertTrue(mail['reply-to'] == 'xyz@gmail.com' )
+
+
```
temparus, line 257:
```suggestion

```

amivapi/utils.py
```diff
@@ -74,7 +74,7 @@ def get_id(item):
         return ObjectId(item['_id'])
 
 
-def mail(to, subject, text):
+def mail(to, subject, text, replyTo = None):
```
temparus, line 77: I suggest to rename the argument to `reply_to` to conform with PEP8 code style.

You should also describe the additional argument in the documentation in the docstring above.

amivapi/utils.py
```diff
@@ -90,17 +90,28 @@ def mail(to, subject, text):
...
+                'reply-to': replyTo
+            })
+        else: 
+          app.test_mails.append({
             'subject': subject,
             'from': sender,
             'receivers': to,
             'text': text
-        })
+        })
```
temparus, line 107:
```suggestion
        mail = {
            'subject': subject,
            'from': sender,
            'receivers': to,
            'text': text,
        }
        
        if replyTo is not None:
            mail['reply-to'] = replyTo

        app.test_mails.append(mail)
```

amivapi/utils.py
```diff
@@ -90,17 +90,28 @@ def mail(to, subject, text):
...
             'text': text
-        })
+        })  
     elif config.SMTP_SERVER and config.SMTP_PORT:
         msg = MIMEText(text)
         msg['Subject'] = subject
         msg['From'] = sender
         msg['To'] = ';'.join([to] if isinstance(to, str) else to)
+        if replyTo is not None:
+            msg['reply-to'] = replyTo
```
temparus, line 114:
```suggestion
            msg['Reply-To'] = replyTo
```

Email headers are generally written with captital letters.

temparus (dismissed):
First of all, thank you for contributing to the project!

I directly suggested some minor changes to the code which you can directly apply in this PR if you are happy with the suggested changes.
There are some more things I have commented on which require changes on multiple locations.

If you have any questions, feel free to answer to my comments.

temparus:
As another thought to consider for this feature:

We have plans to introduce a real noreply email address (noreply@amiv.ethz.ch) for sending emails from automated tools. With the current implementation of the reply-to feature, this results in a situation where the user cannot reply to the received email if the event hasn't a moderator set. We should consider to configure a default reply-to email address which is then used if it is configured. If there is no default configured, the API should skip the `Reply-To` email header as the current implementation does already.

Introducing a real noreply email address might also have other complications for example with the blacklist resource which must be considered, but I think that this should be addressed in another PR.

amivapi/tests/events/test_emails.py
```diff
@@ -229,3 +229,55 @@ def test_no_nones_in_emails(self):
         mail = self.app.test_mails[1]
         for field in mail.values():
             self.assertTrue('None' not in field)
+
+    def test_moderator_reply_to(self):
+        # Check whether `reply-to` header is the moderator in email if set.
```
d-portenier, line 234: I think we should use multiline comments here, because docstring (although I'm not sure weather we use them in this project)

amivapi/tests/events/test_emails.py
```diff
@@ -229,3 +229,55 @@ def test_no_nones_in_emails(self):
...
+        self.api.post('/eventsignups', data={
+            'user': str(user['_id']),
+            'event': str(event['_id'])
+        }, status_code=201).json
+
+        mail = self.app.test_mails[0]
+        self.assertTrue(mail['reply-to'] == 'xyz@gmail.com')
+
+    def test_all_emails_for_reply_to_header(self):
+        # Check that the `reply-to` header in all email is set to default
```
d-portenier, line 250: same here

amivapi/tests/events/test_emails.py
NotSpecial: Agreed, we definitely should.

amivapi/tests/events/test_emails.py
temparus: This was addressed with the latest commits.

amivapi/tests/events/test_emails.py
temparus: This was addressed with the latest commits.

lujobi (approved):

PR #441, opened 2022-05-10
Feature/custom signup email text

emorbach:


amivapi/events/model.py
```diff
@@ -436,6 +436,12 @@
                 'json_schema': True,
                 'only_if_not_null': 'spots',
             },
+            'moderator_signup_text': {
```
temparus, line 439: I think the naming of the field is not accurate. It does not matter whether there is a moderator set for the event.

I would rename the field to something like `signup_additional_info`.
Other suggestions for the field name are welcome as I am not sure that this is the best name.

amivapi/settings.py
```diff
@@ -120,6 +120,7 @@
     'event please click this link to free your spot for someone else:\n'
     '\n{link}\n'
     'You cannot sign out of this event after {deadline}.\n\n'
+    '{moderatortext}\n\n'
```
temparus, line 123: I would rename the field to `moderator_text` which makes it more readable.

amivapi/tests/events/test_emails.py
```diff
@@ -229,3 +229,18 @@ def test_no_nones_in_emails(self):
         mail = self.app.test_mails[1]
         for field in mail.values():
             self.assertTrue('None' not in field)
+
+
```
temparus, line 233:
```suggestion

```
Remove one empty line so it is conform with the code style.

amivapi/tests/events/test_emails.py
```diff
@@ -229,3 +229,18 @@ def test_no_nones_in_emails(self):
...
+        user = self.new_object('users')
+
+        signup = self.api.post('/eventsignups', data={
+            'user': str(user['_id']),
+            'event': str(event['_id'])
+        }, status_code=201).json
+
+        mail = self.app.test_mails[0]
+
+        self.assertTrue(text not in mail)
```
temparus, line 246: This line only checks whether the mail dict has a key with the message stored in `text`.

It should rather be something like:

```suggestion
        self.assertTrue(text not in mail['text'])
```

amivapi/events/model.py
```diff
@@ -436,6 +436,12 @@
...
+                "description": "Custom Text from the Moderator that will be sent with the Signup-Email",
```
temparus, line 440: The description should also be updated accordingly. You should also use more precise words.

For example:

```suggestion
                "description": "Additional information from the organizers that will be sent with the signup-accepted email.",
```

temparus (changes requested):
Thanks for your contribution!

The general idea and structure of the code looks good.
I do not like the name of the new field as it might be misleading, so we should rename it to something else such as `signup_additional_info`. I am not sure about that name either, so other suggestions are welcome.

temparus (changes requested):
Thanks for your work! The changes look good.
I have only one comment about the configuration of the newly introduced field which might have to be changed.

temparus:
@emorbach any updates on this PR?

_Please note that we have updated your branch to the latest master which uses Python 3.10 by default._

amivapi/settings.py
```diff
@@ -127,6 +127,7 @@
     'event please click this link to free your spot for someone else:\n'
     '\n{link}\n'
     'You cannot sign out of this event after {deadline}.\n\n'
+    '{additional_info_text}\n\n'
```
temparus, line 130: This template might result in a very large distance between the deadline and the "Best Regards" phrase. We should somehow ensure that `\n\n` after `{additinal_info_text}` is not added if the `additional_info_text` is an empty string.

The following options might be sensible:
- we provide two different templates for those cases.
- we use the templating engine included with flask (eve is based on flask). It is easy to use as seen in one of our other projects [1].

Aside from that, we might think about providing HTML content in addition to the plaintext version so that the email looks more fancy. But this should be handled in a separate PR to minimize the changes induced by a single PR, which makes it easier to review.

[1] https://gitlab.ethz.ch/amiv/kontakt/website-backend/-/blob/master/kontaktweb/mail.py?ref_type=heads#L23

temparus (changes requested):

alexander-schoch-linuxdays (approved):

d-portenier (approved):

alexander-schoch-linuxdays:
This MR contains two more or less distinct changes:
- An additional field `moderator_signup_text` can be set. This is text any participant will receive via email as soon as they are accepted to an event. This is useful for information the participants should know as soon as possible.
- Emails from the API were, until now, pure text in english. This MR migrates to `jinja2` with templates for each e-mail type with a base template, such that all mails by the AMIVAPI look nice. Additionally, every mail is sent as html and plain text, such that the mail client can decide what to display.

PR #442, opened 2022-05-11
Add GitHub Action workflows

### LINKED ISSUE #438: Replace Travis CI
Services of Travis CI for open-source projects (travis-ci.org) is ceased and the services on travis-ci.com require a paid plan.

Therefore, we have to find another service to run our tests for the project. First, viable services have to be identified and evaluated.

temparus:
Add GitHub Action workflows as replacement for Travis CI.

Resolves #438.


.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,117 @@
+name: Pipeline
+
+on:
+  schedule:
+    - cron: '5 4 * * 1'
```
d-portenier, line 5: 5 past 4 on Mondays seems like a somewhat random time
What do you think about using @weekly (Midnight from Sun to Mon) to improve readability?

d-portenier (approved):

.github/workflows/pipeline.yml
temparus: The idea is not to have the cron job scheduled right at the start of an hour as it will usually be delayed because of high load of GitHub servers according to [their documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule).
The time instant when the cron task is being run does not really matter I think, so we can set it to any time we want, but it should preferably be during night time in Zürich.

They do not support the non-standard @weekly synthax, so we need to convert the standard format like `0 0 * * 1`.

PR #461, opened 2022-05-17
Add support for Python 3.10

temparus:
Add support for Python version 3.10 and set it as the default version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7', '3.9']
+        python-version: ['3.7', '3.10']
 
     services:
       mongodb:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.9-alpine
+FROM python:3.10-alpine
 
 # Create user with home directory and no password and change workdir
 RUN adduser -Dh /api amivapi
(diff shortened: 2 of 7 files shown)


temparus:
Blocked for now as (eve-swagger)[https://github.com/pyeve/eve-swagger] is not compatible with Python 3.10.

temparus:
We had to update `eve-swagger` to the latest version to make the amivapi compatible with Python 3.10.
Some minor adjustments were still needed, so the repo was forked to https://github.com/amiv-eth/eve-swagger.

The new eve-swagger builds an OpenAPI 3 document instead of the older Swagger 2.0 document. I had to change the structures for some parts of the documentation creation files.

With the newer eve-swagger version, the response schema for the list endpoints are now correct as show in the following screenshot.

new_list_response

Additionally, this PR introduces correct response bodies for the error responses as seen in this screenshot.

new_error_response

temparus:
Quick remark: The CI job `test (3.9)` does not exist anymore in the pipeline as we just test the newest and the oldest supported Python version. But it is still marked as required in the repository settings because that is the current state in the master branch. 
So before we can merge this PR, we have to change those settings.

This PR is still waiting for a review.

lujobi (approved):
Thanks a lot for you work! Have there also been change on our fork of the eve-swagger project?

temparus:
> Have there also been change on our fork of the eve-swagger project?

We dropped most changes of the previous fork. Instead, we forked the latest version of Eve Swagger, where we have added the following changes:

- Change type of page field from `string` to `integer`
- Remove security scheme (as no tool relies on that) and the authentication is explained separately already
- Remove default error response because every response code is explained explicitly.

PR #464, opened 2022-05-25
Ldap login without sync

### LINKED ISSUE #436: LDAP sync fails if user is not enrolled
If a user with an n.ethz account is not enrolled, some fields in the LDAP response are missing.

The amivapi crashes with an error 500 and prevents those users from logging in.

temparus:
If some LDAP fields are not present in the LDAP response of ETH, the whole user synchronization fails currently and the users cannot log in as described in #436 (Closes #436).

This PR ensures that the users can still successfully log in even if the user synchronization fails but the initial authentication with the LDAP server was successful.

The data processing of the LDAP response is made more failure tolerant. If a field which we expect to exist in the LDAP response is not available, we just skip processing that specific field but still handle the other fields. This makes it a bit harder to spot and debug potential issues with the LDAP sync.

If we want to make that more transparent, we could add further logging messages pointing out which fields were skipped.

### DIFF
--- a/amivapi/auth/sessions.py
+++ b/amivapi/auth/sessions.py
@@ -182,10 +182,10 @@ def process_login(items):
             # Success, sync user and get token
             try:
                 user = ldap.sync_one(username)
+                app.logger.info(
+                    "User '%s' was authenticated with LDAP" % username)
             except LDAPException:
                 # Sync failed! Try to find user in db.
-                app.logger.error(f"User '{username}' was authenticated, " +
-                                 "but could not be synced.")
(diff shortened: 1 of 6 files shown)


temparus:
@NotSpecial an additional test case is now added. The PR is complete and waiting for a review.

temparus:
In the mean time, I am also one of the users affected by the login issue. This PR is ready for another review.

lujobi (approved):

lujobi:
Sorry for the long response, took me another ping via DM 🙈

PR #514, opened 2022-12-17
Fix issue with CI pipeline

temparus:
Ubuntu-22.04 has no out-of-the-box support for Mongo 5. Therefore we keep using ubuntu-20.04.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   test:
-    # Due to incompatibility of ubuntu-20.04 with mongo 5.0,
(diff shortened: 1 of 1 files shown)


suisseWalter (approved):
looks good to me.

alexander-schoch-linuxdays (approved):

.github/workflows/ci.yml
```diff
@@ -9,6 +9,8 @@ on:
 
 jobs:
   test:
+    # Due to incompatibility of ubuntu-20.04 with mongo 5.0,
```
alexander-schoch-linuxdays, line 12: shouldn't it be 22.04?

temparus: Indeed. It's fixed now.

PR #515, opened 2022-12-17
Drop support for Python 3.7

temparus:
New version of some dependencies do not support Python 3.7 anymore. As we use a newer Python version anyway, we should drop this support too.


suisseWalter (approved):
LGTM, Can't find any other cases of the version being mentioned.

PR #517, opened 2022-12-18
Add `time_deregister_end` field to events resource

temparus:
The requirement emerged to have different signup and signoff deadlines, so that users can still signoff when users cannot signoff themselves anymore.

**Important Note**
The new field is required if a registration window is given. This means that the existing data has to be migrated when deploying this feature.

### DIFF
--- a/amivapi/tests/fixtures.py
+++ b/amivapi/tests/fixtures.py
@@ -258,12 +258,17 @@ class FixtureMixin(object):
                 obj['time_register_end'] = (
                     datetime.utcnow() + timedelta(
                         seconds=random.randint(0, 1000000)))
+                obj['time_deregister_end'] = (
+                    datetime.utcnow() + timedelta(
+                        seconds=random.randint(0, 1000000)))
             else:
                 if ('time_register_start' not in obj or
-                        'time_register_end' not in obj):
+                        'time_register_end' not in obj or
+                        'time_deregister_end' not in obj):
                     raise BadFixtureException(
-                        "Bad fixture: please specify either both "
-                        "time_register_start and time_register_end or none")
+                        "Bad fixture: please specify either all of "
+                        "time_register_start and time_register_end "
+                        "and time_deregister_end or none")
 
             obj.setdefault('allow_email_signup',
                            random.choice([True, False]))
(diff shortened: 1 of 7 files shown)


amivapi/tests/fixtures.py
```diff
@@ -258,12 +258,17 @@ def preprocess_events(self, schema, obj, fixture):
                 obj['time_register_end'] = (
                     datetime.utcnow() + timedelta(
                         seconds=random.randint(0, 1000000)))
+                obj['time_deregister_end'] = (
+                    datetime.utcnow() + timedelta(
+                        seconds=random.randint(0, 1000000)))
```
d-portenier, line 263: time_deregister_end should be <= time_registter_end
Otherwise, people can still sign off after they can sign up which contradicts the use case for which we implement this feature.

amivapi/tests/fixtures.py
```diff
@@ -258,12 +258,17 @@ def preprocess_events(self, schema, obj, fixture):
...
             else:
                 if ('time_register_start' not in obj or
-                        'time_register_end' not in obj):
+                        'time_register_end' not in obj or
+                        'time_deregister_end' not in obj):
                     raise BadFixtureException(
-                        "Bad fixture: please specify either both "
-                        "time_register_start and time_register_end or none")
+                        "Bad fixture: please specify either all of "
+                        "time_register_start and time_register_end "
```
d-portenier, line 270:
```suggestion
                        "time_register_start, time_register_end "
```

amivapi/tests/fixtures.py
d-portenier: Also move deregister_end on top of register_end, as this deregister_end is (in general) earlier than register_end

test_performance.py
```diff
@@ -102,6 +102,8 @@ def create_event():
         'time_register_start': datetime(1970, 1, 1).strftime(DATE_FORMAT),
         'time_register_end': (datetime.utcnow() +
                               timedelta(days=100)).strftime(DATE_FORMAT),
+        'time_deregister_end': (datetime.utcnow() +
+                              timedelta(days=100)).strftime(DATE_FORMAT),
```
d-portenier, line 106: Also move deregister_end in frnt of register_end

alexander-schoch-linuxdays (approved):

PR #575, opened 2023-07-20
made joboffers .jpg compatible

ahirsch-dev:
test also expanded


temparus (approved):
Thanks for your first contribution to the project!

Very nice that you have also updated the unittests :)

PR #577, opened 2023-07-23
fix bug that the english weekday was taken for both languages; include deregistration time

alexander-schoch-linuxdays:


d-portenier:
Consider en/de instead of german/english for consistency with rest of repo.

d-portenier (approved):

PR #602, opened 2023-10-14
Allow grouping of dependabot PRs

temparus:
Update the dependabot configuration so that certain updates are grouped together in a single PR.


suisseWalter (approved):
this is fine, only affects the frequency for the pipeline. the rest is just grouping.

Issue #493, opened 2022-09-15
eventsignups: no confirmation-email sent after manual accept


niccantieni:
It seems like no confirmation-email is sent after a user was manually accepted to an event (i.e. the mode is `manual`, not `fcfs`). 
The issue was noticed on the event `62b32980547064b604c46645`,  [link](https://amiv.ethz.ch/de/events/62b32980547064b604c46645). 

I signed up and was placed on the waiting list immediately (which is correct, since the manual mode was selected),. Then, at some point in the last days I was manually accepted, but never got a confirmation email. 

It seems like it was intended to add this confirmation email in PR #159, but I cannot tell whether the code is actually correct. Maybe the code only works if one is promoted from a waiting list which exists because the event is full (and not because it is a manual event)?
Some of the relevant code is [here](https://github.com/amiv-eth/amivapi/blob/c7aaecf5b680756bcaa5397a5774a48617839157/amivapi/events/queue.py#L73-L86).

Possible other issue: maybe the request sent by the admin-tool is wrong, i.e. when the "accept"-button in the waiting-list table is pressed, the request is missing some values? Maybe something goes wrong with the explicit `accept: true` required by the API, which can only be set by the API itself or by moderators (according to [the documentation](https://api.amiv.ethz.ch/docs#tag/Event-Signup/paths/~1eventsignups~1%7Bevent%20signupId%7D/patch)).
The request is assembled [here](https://gitlab.ethz.ch/amiv/amiv-admintool/-/blob/master/src/events/participants.js#L138-L157).

I am going to look into this some more, but any pointers as to what may be wrong are much appreciated.

niccantieni:
As an afterthought: this is probably related to the work done in #416.

kunerj:
Is it possible that manual selection mode has rarely been used and is thus not really field-tested?

Looking at the [hooks defined for eventsignups ](https://github.com/amiv-eth/amivapi/blob/a5c11471298a7f51d60117d37eb46d7408942cfd/amivapi/events/__init__.py#L49-L73), there is no one for update/patch requests to the eventsignups [sent by the frontend](https://gitlab.ethz.ch/amiv/amiv-admintool/-/blob/master/src/events/participants.js#L138-L157) you mentioned. I would have expected a `on_updated_eventsignups` hook, or am I missing something @temparus ?

kunerj:
Fixed in #605

PR #605, opened 2023-10-15
Issue 493: Fix missing email notifications on manual waitlist updates

kunerj:
Tackling issue #493 of missing notifications if eventsignups items are updated via PATCH requests. The corresponding hook was missing.


suisseWalter (approved):
This looks good, it should now cover further cases, which should now be all possible. 
case2 should rarely happen

PR #611, opened 2023-11-18
Enhance Authentication

### LINKED ISSUE #598: Login with email address
Truncate any emailadress that matches `*@*ethz.ch` to everything before the `@`.

temparus:
- Fixes the LDAP sync issue for users which are not enrolled at ETH anymore.
- Allows for ETH members to authenticate against LDAP by using their ETH email address (Resolves #598)

### DIFF
--- a/amivapi/auth/sessions.py
+++ b/amivapi/auth/sessions.py
@@ -5,6 +5,7 @@
 """Sessions endpoint."""
 
 import datetime
+import re
 
 from amivapi import ldap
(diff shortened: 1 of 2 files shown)


suisseWalter (approved):
lgtm, doesn't change the behaviour for extraordinary members hat sign in with their private email, but allows eth students to sign in with their full email address. which is now common. also fixes the the issue with signins for extraordinary members, that haven't changed their email/password.

PR #624, opened 2024-01-02
Rewrite Waiting List Email

alexander-schoch-linuxdays:
There's two cases why a user would receive the "events_waitignList" email:

1. The event is full. In this case, they end up on the waiting list and can be accepted if people sign off.
2. The event moderator has to approve of the participants, and thus all users end up on the waiting list.

The current text reads as follows:

```
Hello {{ name }}!

We are sorry to inform you that your signup for {{ event }} was rejected because there are no free spots available.

We put you on a waiting list and we will send you an email when a spot becomes available.

Best Regards,
AMIV
```

This text is well-suited for case 1, but very confusing for case 2. I got confused twice within one month about this now, so I'm creating a PR.

### DIFF
--- a/amivapi/templates/events_waitingList.html
+++ b/amivapi/templates/events_waitingList.html
@@ -3,8 +3,8 @@
 {% block text_german %}
   <p>Hey {{ name }}!</p>
 
-  <p>Leider müssten wir dich informieren, dass deine Anmeldung für {{ title_de }} abgelehnt wurde, weil keine freien Plätze mehr verfügbar sind.</p>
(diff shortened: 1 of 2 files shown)


suisseWalter:
looks good. This has been a criticism for a long time. and this is a nice solution that doesn't require change any logic, the suggestion was always that there should be two different emails.

nimro27 (approved):
Looks good!

temparus (approved):

temparus:
As we check for specific phrases in our unit tests, you should update the test cases in `amivapi.tests.events.test_emails.py`.

PR #666, opened 2024-09-14
python 12 dev

suisseWalter:
update to py3.12 should bring performance and maybe stability.
python 3.8 is EOL, therefore we also remove it. 
decreased agressiveness of codecov as it wouldn't let this PR through even though no actual line of code has changed.


nimro27:
Looks good. Merging the changes

nimro27 (approved):

PR #669, opened 2024-10-21
Send a different email for waitinglist on manually accepted events

michaelstambach:
Send a different email if a person is put on the waitinglist for an event with manual acceptance that still has free spots.
This essentially fixes the same problem as #624 but with different emails, which I only noticed after I had already implemented the changes.


nimro27:
Looks good.

nimro27 (approved):
Looks good. Thanks!

